PR #6, opened 2015-08-15
C#/XML syntax highlighting

sergey-shandar:
Syntax highlighting in readme.md file.


AArnott:
Thanks. Thus far I avoided using a lot of github/markdown specific syntax because this same readme is used when the package is installed in VS and VS doesn't render markdown particularly so I wanted to keep it readable. Forking the readme between the two purposes can increase readability for both but increases maintenance. I'm not sure what to do yet.

sergey-shandar:
Interesting. I didn't know that it is not standard feature. I know that GitHib, VSO and Visual Studio Code support the syntax but
- StackOverflow has it's own syntax highlighting such as: 
- MarkDownPad doesn't support the highlighting syntax.

We need VS support for markdown.

PR #13, opened 2015-09-24
Enable build version override

### LINKED ISSUE #15: BuildNumber property should be set unconditionally
The BuildNumber project property is set by TFS Team Build, and thus this project's idea for BuildNumber isn't applied because it is only set by this project if the property is empty beforehand. We should remove the condition.

adrianvmsft:
Fixes #15


src/Nerdbank.GitVersioning.NuGet/build/NerdBank.GitVersioning.targets
```diff
@@ -34,7 +34,7 @@
       <Output TaskParameter="BuildNumber" PropertyName="BuildVersionNumberComponent" />
     </GetBuildVersion>
     <PropertyGroup>
-      <BuildNumber Condition=" '$(BuildNumber)' == '' ">$(BuildVersionNumberComponent)</BuildNumber>
+      <BuildNumber Condition=" '$(BuildNumber)' == '' or '$(OverrideBuildNumber)' == 'true' ">$(BuildVersionNumberComponent)</BuildNumber>
```
AArnott: If the caller is going to have to add a property definition to allow the BuildNumber to be reset, why not ask them to simply clear BuildNumber rather than having to set OverrideBuildNumber=true? It's one less property to document.

adrianvmsft: In some cases, this gets passed in by the build definition. In some cases controlling that may not be feasible.

AArnott: AFAIK properties are either immutable (e.g. toolset properties, or when set as a global property at the command line) or mutable (environment variables, project properties). Immutable ones wouldn't be fixed by this. Mutable ones can always be cleared. Build definitions can only pass in the properties using an inserted .targets files, environment variables or global properties. 

If it's an inserted .targets file, and that .targets file is inserted just above where gitversioning.targets is imported, I could see that being a problem to find a good place to clear the property. But I suspect these TFS builds are just setting it as an environment variable. Have you tried clearing it in EnlistmentInfo.props?

AArnott: Alright, so after much flip flopping in my mind, I think the best thing to do is remove the condition altogether. We don't use that condition anywhere else, and it *is* the purpose of using this package to determine the version after all, so by all means, let's do it. :) Can you update your pull request?

Issue #12, opened 2015-09-22
Automate the generation of version.txt and removal of assembly attributes when the package is installed


jkeech:
Remove the need to manually create a version.txt file and update AssemblyInfo.cs after the NuGet package is installed. By leveraging NuGet's Install.ps1 script, we should be able to automate this process by seeding the version.txt file with the Major.Minor from the AssemblyVersion attribute in AssemblyInfo.cs.

sergey-shandar:
The problem with installation script is it runs every time NuGet is restored/updated etc. but a user expect this changes happen only once. Another problem with this, if someone doesn't like the tool and would like to uninstall it, uninstall script will not be able to revert the changes.

jkeech:
It's not a problem, you just need to design the script so that it only creates the version.txt file (and removes assembly attributes) if a version.txt file doesn't already exist, that way it only runs on first install and no-ops during upgrades or if the user already had created the version.txt file manually.

Likewise, you can include an Uninstall.ps1 script which can reverse the vast majority of the logic: Delete version.txt, then add the AssemblyVersion attribute back into AssemblyInfo.cs using the Major.Minor from version.txt. This solves the 99% case. The other attributes (AssemblyFileVersion, AssemblyInformationalVersion) would need to be updated manually anyways after disabling this package, because they will have different file/informational versions and they need to ensure that the next version they produce is higher than the previous versions.

Your argument against automating these configuration steps is basically that you don't want NuGet to be modifying the project on behalf of the user, but that is one of the core features of NuGet and is well supported and expected behavior when installing packages. I think the benefit of making the package as easy to use as possible outweighs the downsides here. With the install script, I don't have to do any configuration whatsoever. I just install the package, commit the automated changes, and build.

AArnott:
@jkeech said:
> our argument against automating these configuration steps is basically that you don't want NuGet to be modifying the project on behalf of the user, but that is one of the core features of NuGet and is well supported and expected behavior when installing packages. 

Actually, NuGet v3 deprecates .ps1 scripts and in the project.json world they're not supported at all. Ps1 scripts are infamous for not being adequately tested or working in all scenarios so they don't like them and discourage their use. So adding this at this point seems a little like swimming upstream. I do like having scripts that can automate it though, as this is typically a tedious process. And it does no harm to have the .ps1 script in the project.json world (they'll just be quietly ignored).

AArnott:
[...]

PR #16, opened 2015-09-24
Add an Install.ps1 script to automate version configuration

jkeech:
Implements #12

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -16,10 +16,26 @@ What sets this package apart from other git-based versioning projects is:
 3. No dependency on branch names. Branches come and go, and a commit may belong to any number of branches. Regardless of the branch HEAD may be attached to, the build should be identical.
 4. The computed version information is based on an author-defined major.minor version and an optional unstable tag, plus a shortened git commit ID.
 
-## Installation
+## Installation and Configuration
 
-After installing this NuGet package, you may find that you get a compilation failure
-because of multiple definitions of certain attributes such as `AssemblyVersionAttribute`.
+After installing this NuGet package, you may need to configure the version generation logic
+in order for it to work properly.
+
+With NuGet 2.x, the configuration is handled automatically via the tools\Install.ps1 script.
+For NuGet 3.x, you can run the script tools\Create-VesionTxt.ps1 to help you create the
+version.txt file and remove the old assembly attributes.
+
+The scripts will look for the presence of a version.txt file. If it already exists, nothing
+happens. If the version.txt file does not exist, the script looks in your project for the
+Properties\AssemblyInfo.cs file and attempts to read the Major.Minor version number from
+the AssemblyVersion attribute. It then generates a version.txt file using the Major.Minor
+that was parsed so that your assembly will build with the same AssemblyVersion as before,
+which preserves backwards compatability. Finally, it will remove the various version-related
+assembly attributes from AssemblyInfo.cs.
+
+If you did not use the scripts to configure the package, you may find that you get a
+compilation failure because of multiple definitions of certain attributes such as
+`AssemblyVersionAttribute`.
 You should resolve these compilation errors by removing these attributes from your own
 source code, as commonly found in your `Properties\AssemblyInfo.cs` file:
 
@@ -30,10 +46,19 @@ source code, as commonly found in your `Properties\AssemblyInfo.cs` file:
 This NuGet package creates these attributes at build time based on version information
 found in your `version.txt` file and your git repo's HEAD position.
 
+Note: After first installing the package, you need to commit the version.txt file so that
+it will be picked up during the build's version generation. If you build prior to committing,
+the version number produced will be 0.0.x.
+
 ### version.txt file
 
 You must define a version.txt file in your project directory or some ancestor of it.
-By convention it is often found in the root directory of your git repo.
+
+When the package is installed, a version.txt file is created in your project directory
+(for NuGet 2.x clients). This ensures backwards compatability where the installation of
+this package will not cause the assembly version of the project to change. If you would
+like the same version number to be applied to all projects in the repo, then you may move
+the file to the root directory of your git repo.
 
 Here is the content of a sample version.txt file you may start with (do not indent):
 
--- a/readme.md
+++ b/readme.md
@@ -22,7 +22,7 @@ After installing this NuGet package, you may need to configure the version gener
 in order for it to work properly.
 
 With NuGet 2.x, the configuration is handled automatically via the tools\Install.ps1 script.
-For NuGet 3.x, you can run the script tools\Create-VesionTxt.ps1 to help you create the
+For NuGet 3.x, you can run the script tools\Create-VersionTxt.ps1 to help you create the
 version.txt file and remove the old assembly attributes.
 
 The scripts will look for the presence of a version.txt file. If it already exists, nothing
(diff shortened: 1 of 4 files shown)


AArnott:
Thanks. I look forward to reviewing this.

readme.md
```diff
@@ -16,10 +16,26 @@ What sets this package apart from other git-based versioning projects is:
...
-
-After installing this NuGet package, you may find that you get a compilation failure
-because of multiple definitions of certain attributes such as `AssemblyVersionAttribute`.
+## Installation and Configuration
+
+After installing this NuGet package, you may need to configure the version generation logic
+in order for it to work properly.
+
+With NuGet 2.x, the configuration is handled automatically via the tools\Install.ps1 script.
+For NuGet 3.x, you can run the script tools\Create-VesionTxt.ps1 to help you create the
```
AArnott: I think it's more based on whether it's packages.config vs. project.json than on the version of NuGet itself.
And you are missing an 'r' in Create-VesionTxt.ps1

AArnott:
I like how you factored the meat of it in a powershell script of its own. :)

readme.md
```diff
@@ -16,10 +16,26 @@ What sets this package apart from other git-based versioning projects is:
...
+version.txt file and remove the old assembly attributes.
+
+The scripts will look for the presence of a version.txt file. If it already exists, nothing
+happens. If the version.txt file does not exist, the script looks in your project for the
```
AArnott: That suggests that installing the package into the second project of my solution will notice the original version.txt file, and not clear out the assemblyinfo.cs file. Perhaps we should always check the assemblyinfo.cs file for version attributes and remove them?

readme.md
```diff
@@ -16,10 +16,26 @@ What sets this package apart from other git-based versioning projects is:
...
+Properties\AssemblyInfo.cs file and attempts to read the Major.Minor version number from
+the AssemblyVersion attribute. It then generates a version.txt file using the Major.Minor
+that was parsed so that your assembly will build with the same AssemblyVersion as before,
+which preserves backwards compatability. Finally, it will remove the various version-related
```
AArnott: sp: compatability->compatibility

readme.md
```diff
@@ -16,10 +16,26 @@ What sets this package apart from other git-based versioning projects is:
...
+in order for it to work properly.
+
+With NuGet 2.x, the configuration is handled automatically via the tools\Install.ps1 script.
+For NuGet 3.x, you can run the script tools\Create-VesionTxt.ps1 to help you create the
+version.txt file and remove the old assembly attributes.
+
+The scripts will look for the presence of a version.txt file. If it already exists, nothing
+happens. If the version.txt file does not exist, the script looks in your project for the
+Properties\AssemblyInfo.cs file and attempts to read the Major.Minor version number from
+the AssemblyVersion attribute. It then generates a version.txt file using the Major.Minor
```
AArnott: Can you specify where the version.txt file will be placed?
I hope it can be at the solution directory, and thus shared with other projects in the solution by default.

readme.md
```diff
@@ -30,10 +46,19 @@ source code, as commonly found in your `Properties\AssemblyInfo.cs` file:
...
+it will be picked up during the build's version generation. If you build prior to committing,
+the version number produced will be 0.0.x.
+
 ### version.txt file
 
 You must define a version.txt file in your project directory or some ancestor of it.
-By convention it is often found in the root directory of your git repo.
+
+When the package is installed, a version.txt file is created in your project directory
+(for NuGet 2.x clients). This ensures backwards compatability where the installation of
```
AArnott: sp: compatability -> compatibility

src/Nerdbank.GitVersioning.NuGet/tools/Create-VersionTxt.ps1
```diff
@@ -0,0 +1,71 @@
...
+            {
+                # Grab the Major.Minor out of this file which will be injected into the version.txt
+                $version = $matches[1] + ".0"
+            }
+
+            # Remove attributes related to assembly versioning since those are generated on the fly during the build
+            $_ -replace "^\[assembly: Assembly(?:File|Informational|)Version\(""[0-9]+(?:.(?:[0-9]+|\*)){1,3}""\)\]$"
+        }
+
+        if ($PSCmdlet.ShouldProcess($assemblyInfo, "Removing assembly attributes"))
```
AArnott: Love it!

AArnott:
Looks really good overall. It doesn't work (automatically) in VS2015 with packages.config, presumably because it's NUget v3. I'm spinning up a VS2013 VM to test it. But once the issues above are resolved and I've tested that it works myself, I'll merge.

jkeech:
Hmm, I also tested with VS 2015. What was the outcome vs what you expected?

readme.md
jkeech: I'm a bit torn on the scope issue. It creates a version.txt at the project level for safety (back-compat). Say you have two projects in a solution that are currently building different assembly versions prior to adding Nerdbank.GitVersioning to both projects. Due to how assembly versions work, I wouldn't want this package to automatically change my assembly version because other people might have compiled against and older version of my code already. So the safest thing to do is to have the package always generate the same assembly version as before the package was installed. In this case, you must generate version.txt at the project level. But I agree that in the common use case, you'll want solution- or repo-level consistent versioning with the a common version across a set of multiple packages. So there's a tradeoff between safety and simplicity.

We could make it generate version.txt at the repo or solution level, and then when the package is installed in other projects, it could just remove the assembly attributes and be done. Since it works at the project level right now, it only does the assemblyinfo check if version.txt doesn't exist (because if the file did exist, the package was already installed before or the user has manually created version.txt at the project level and probably also cleaned up assemblyinfo.cs already).

AArnott: Fair point. OK.

AArnott:
I hoped it would work on all versions of vs and nuget as long it was packages.config but if that doesn't work I suppose that's just a nuget limitation and not a bug on your part. I'll double check tomorrow.

readme.md
jkeech: I do think there's a slight improvement I could make though: If version.txt does not exist and the assemblyversion attribute also does not exist in assemblyinfo.cs, than do not generate a version.txt at the project level, since this project is most likely intentionally consuming a version.txt from the parent already. This would enable a cleaner upgrade scenario where version.txt would not be added in at the project level during an upgrade if it was previously deleted/moved from the project directory.

jkeech:
FWIW, I tested with VS 2015 Community and NuGet Package Manager for Visual Studio version 3.2.60914.10516.

To test, I created a new console app in a repo, and put some number in the AssemblyVersion attribute, such as "2.3.0.0". I committed, and then installed the package with my changes via right-click -> "Manage NuGet Packages..." on the project which is using packages.config. What resulted was a new version.txt file added to the project with contents "2.3.0", and the AssemblyVersion and AssemblyFileVersion lines removed from AssemblyInfo.cs. I then committed these changes so that version.txt is in the git history and built the project. It resulted in version 2.3.1.46962.

AArnott:
For some really weird reason, the .nupkg your PR builds has a tools folder yet when VS downloads it, it doesn't. That's why the ps1 script isn't executing for me. I'm still investigating.

AArnott:
OK, so AppVeyor is pulling some crazy stunt and has modified the NuPkg file before publishing it to its feed to remove the install.ps1 script. See the discussion I've started here: http://help.appveyor.com/discussions/problems/3122-nuget-package-from-feed-is-missing-files-that-are-present-in-build-artifacts

PR #39, opened 2015-12-23
Support automatically building as if PublicRelease=true

### LINKED ISSUE #38: Build server support: build certain tags as PublicRelease=true
When branches and tags are pushed, builds are sometimes automatically scheduled on build servers. For example on AppVeyor, pushing [build tags kicks off a build with the `APPVEYOR_REPO_TAG_NAME` environment variable preset](http://www.appveyor.com/docs/environment-variables).

We can add a `releaseRefRegex` field to the version.json file such that when builds are kicked off based on tags or branches that match the regular expression provided then the `PublicRelease` MSBuild property is set to `true` automatically, thus producing a build without the `-gCOMMITID` suffix to the NuGet package versions.

This enables the flow of a build completing and being validated, then deciding to ship that version. The developer then pushes a tag (say... "v1.0.1") and that automatically kicks off a fresh build that is just "1.0.1" instead of "1.0.1-gabc123def".

AArnott:
The version.json file can now specify regex patterns that if the ref being built matches will cause PublicRelease to default to true instead of false.

This comes with specialized support for a couple of build server platforms (VSTS and AppVeyor). But it works with pure git as well, both locally on dev boxes and on these and other build servers.

Fix #38

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -257,7 +257,7 @@ public class BuildIntegrationTests : RepoTestBase
         var versionOptions = new VersionOptions
         {
             Version = SemanticVersion.Parse("1.0"),
-            PublicReleaseRefSpec = "^refs/heads/release$",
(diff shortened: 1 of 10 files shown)


AArnott:
Can we support arrays of regular expressions instead of just one?

PR #46, opened 2016-01-19
Add support for AssemblyTitle/AssemblyProduct/etc. to AssemblyInfo task

azeno:
In previous versions the AssemblyInfo task was taken from https://github.com/loresoft/msbuildtasks and I kinda misused it to also set ```AssemblyTitle```, ```AssemblyProduct```, ```AssemblyCompany``` and ```AssemblyCopyright``` attributes. I was also using the ```ThisAssembly.AssemblyConfiguration``` constant, which I'm missing now in the latest release (1.2.8).

This PR adds all those fields and should hopefully not break anything.

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -359,19 +359,32 @@ public class BuildIntegrationTests : RepoTestBase
 
         var assemblyFileVersion = assembly.GetCustomAttribute<AssemblyFileVersionAttribute>();
         var assemblyInformationalVersion = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>();
+        var assemblyTitle = assembly.GetCustomAttribute<AssemblyTitleAttribute>();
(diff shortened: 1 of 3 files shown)


AArnott:
> I kinda misused it 

:laughing: I'm glad you recognize that. :) I would have won't fixed this if it had been an issue, but seeing as you've gone to the trouble of a PR, I'll review it and do my best to accept it. :)

azeno:
Well I wasn't sure what my alternatives would've been. Both tasks are called ```AssemblyInfo```, both do nearly the same thing, sure I could've set the Company, Copyright and Product fields in the AssemblyInfo.cs files directly, but to derive AssemblyTitle from $(AssemblyName) and AssemblyConfiguration from $(Configuration) was very neat.

AArnott:
Good point. When you put it that way, maybe we should rename our MSbuild task. How do you feel about adding that to the pull request?

azeno:
I chose AssemblyVersionInfo as its main purpose is to set the version of the assembly. That ok?

AArnott:
Looks fantastic!

Thanks.

PR #68, opened 2016-05-24
The GetBuildVersion task crashed in case a project was not in a subdirectory

azeno:
Encountered this issue in a project where the directory layout was like this:
`ROOT_DIR/.git/`
`ROOT_DIR/*.csproj`
`ROOT_DIR/version.json`

When building the project it crashed with the error message
`startindex cannot be larger than length of string`

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -37,6 +37,7 @@ public class BuildIntegrationTests : RepoTestBase
     private ProjectCollection projectCollection;
     private string projectDirectory;
     private ProjectRootElement testProject;
+    private ProjectRootElement testProjectInRoot;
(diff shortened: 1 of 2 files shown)


AArnott:
This is fantastic. Thanks so much for the contribution and in particular the regression test you added.

src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
```diff
@@ -76,7 +78,7 @@ public BuildIntegrationTests(ITestOutputHelper logger)
     {
         this.WriteVersionFile();
         this.InitializeSourceControl();
-        var buildResult = await this.BuildAsync();
+        var buildResult = await this.BuildAsync(this.testProject);
```
AArnott: I think it would help keep the tests simple if we stuck with `this.testProject` being implied, and the one test that needs a specialized project can just reassign `this.testProject` with the rooted project.
I'll build a commit on top of yours to do that and then merge.

PR #69, opened 2016-05-31
Fully qualify build target invocations and declarations

jeremyhaubold:
Also make non-version attribute emission optional

### DIFF
--- a/src/Nerdbank.GitVersioning.NuGet/build/NerdBank.GitVersioning.targets
+++ b/src/Nerdbank.GitVersioning.NuGet/build/NerdBank.GitVersioning.targets
@@ -22,9 +22,9 @@
     <GenerateAssemblyVersionInfo Condition=" '$(GenerateAssemblyVersionInfo)' == '' and '$(TargetExt)' != '.dll' and '$(TargetExt)' != '.exe'">false</GenerateAssemblyVersionInfo>
   </PropertyGroup>
 
-  <UsingTask AssemblyFile="$(NerdbankGitVersioningTasksPath)Nerdbank.GitVersioning.Tasks.dll" TaskName="AssemblyVersionInfo"/>
-  <UsingTask AssemblyFile="$(NerdbankGitVersioningTasksPath)Nerdbank.GitVersioning.Tasks.dll" TaskName="GetBuildVersion"/>
-  <UsingTask AssemblyFile="$(NerdbankGitVersioningTasksPath)Nerdbank.GitVersioning.Tasks.dll" TaskName="CompareFiles"/>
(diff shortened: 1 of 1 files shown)


src/Nerdbank.GitVersioning.NuGet/build/NerdBank.GitVersioning.targets
```diff
@@ -33,9 +33,18 @@
     <_NBGV_BuildingRef>$(_NBGV_BuildingTag)</_NBGV_BuildingRef>
     <_NBGV_BuildingRef Condition=" '$(_NBGV_BuildingRef)' == '' ">$(_NBGV_BuildingBranch)</_NBGV_BuildingRef>
   </PropertyGroup>
-
+  <PropertyGroup Condition=" '$(NBGV_DoNotEmitNonVersionCustomAttributes)' != 'true' ">
+    <_NBGV_AssemblyName>$(AssemblyName)</_NBGV_AssemblyName>
+    <_NBGV_RootNamespace>$(RootNamespace)</_NBGV_RootNamespace>
```
AArnott: Why did you need to provide special treatment to this (and some other) properties? Not all of these become custom assembly attributes. The list of properties that become attributes is [here](https://github.com/AArnott/Nerdbank.GitVersioning/blob/master/src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs#L169). Can you reduce the specially handled set to (a subset of) these?

jeremyhaubold: Sure thing.

AArnott:
Thanks!

jeremyhaubold:
How/when do new packages get published?

AArnott:
It's available right now in this nuget package feed: https://ci.appveyor.com/nuget/nerdbank-gitversioning
package version: 1.4.41

It'll eventually be pushed to nuget.org, of course. It would be great if you could validate that these changes adequately unblock you though. Can you use the nuget package feed above to accomplish that?

jeremyhaubold:
It's not immediately clear how to use that as a feed, but I've validated locally by hotfixing the package with the identical change.

AArnott:
Thanks for the validation.
FYI, you can consume packages from many feeds just by adding a nuget.config file, which [I document on another github repo here](https://github.com/AArnott/pinvoke/blob/master/CONTRIBUTING.md#self-service-releases-for-contributors).

AArnott:
I have pushed the new version to nuget.org

jeremyhaubold:
Yes, I'm familiar with adding multiple feeds, but https://ci.appveyor.com/nuget/nerdbank-gitversioning itself doesn't appear to be a valid feed URL.

AArnott:
I see. Yes, it should work just fine. I don't know whether it's a v2 or v3 package feed (probably v2).

PR #72, opened 2016-06-13
dropped unknown compiler option, added request dependency

vladima:
a few questions:
- do you use `tsd` or `typings`? it is not clear since you have dependency on `typings` but `tsd.json`
- reason of faulty intellisense - unknown compiler option `sortOutput` in `tsconfig.json` - why it was added?

### DIFF
--- a/src/nerdbank-gitversioning.npm/package.json
+++ b/src/nerdbank-gitversioning.npm/package.json
@@ -22,7 +22,8 @@
     "merge2": "1.0.2",
     "q": "1.4.1",
     "typescript": "1.8.10",
-    "typings": "1.0.5"
+    "typings": "1.0.5",
+    "tsd": "^0.6.5"
   },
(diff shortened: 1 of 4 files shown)


AArnott:
I had added `sortOutput` to the tsconfig.json file because it was semi-documented in [gulp-typescript](https://www.npmjs.com/package/gulp-typescript). But I'm happy to remove it.

AArnott:
I hear `typings` is the new thing and `tsd` is the old. So I'm happy to use the newest thing. But I'm very new to the area, so I've used a mixture of things I've seen others use and things I dug up in docs on my own.

AArnott:
Thanks. I had gone on to change some code, including making a couple of the fixes you made, before your PR came in. I've manually made the rest, and switched entirely to typings (tsconfig.json is now gone). 

Thanks again!

PR #73, opened 2016-06-13
Allow specifying Git repo root manually in case build is running with…

jeremyhaubold:
… a working directory that is outside the repo

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -92,6 +92,30 @@ public class BuildIntegrationTests : RepoTestBase
     }
 
     [Fact]
+    public async Task GetBuildVersion_OutsideGit_PointingToGit()
+    {
+        // Write a version file to the 'virtualized' repo.
+        string version = "3.4";
+        this.WriteVersionFile(version);
+
+        // Update the repo path so we create the 'normal' one elsewhere
+        this.RepoPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+        this.InitializeSourceControl();
+
+        // Write the same version file to the 'real' repo
+        this.WriteVersionFile(version);
+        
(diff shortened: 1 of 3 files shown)


AArnott:
When would your scenario happen? The working directory is always set by MSBuild to be the directory containing the project being built. If the project itself isn't in the git repo, why would it be stamped with version info from that repo?

AArnott:
Your changes look just fine, BTW.

jeremyhaubold:
It is a CloudBuild scenario. All the builds happen in virtualized folders that are not necessarily "underneath" the repo root, so the search-upwards logic can't find the repo.

Phone = brevity
________________________________

AArnott:
I see. Then this change sounds reasonable.
However, can we rename `GitRepoRoot` to `ProjectDirectory`? The reason being that the version calculated may vary depending on *where* in the git repo it is due to having version.json in any folder other than the git root. So indicating the logical directory within that git repo can be important, and `GitRepoRoot` doesn't capture that.

jeremyhaubold:
Sure, will update.

Phone = brevity
________________________________

jeremyhaubold:
Updated.

AArnott:
Looking over this some more (sorry, I'm discovering this incrementally) it looks like `Environment.CurrentDirectory` is used in other places too (like finding the version.json I mentioned earlier). Going back to your original scenario that's prompting this change, what should happen there? Is the current working directory still appropriate or should it use the overridden path? 

I guess the path you changed is only relevant for searching for the actual .git folder so in that respect your property name of `GitRepoRoot` makes sense, particularly since you *don't* change the code that searches from the project directory for the version.json file. So if that's actually an intentional focused change, then 😢 your first commit was I guess better than the one I asked you to change it to. Do you still have that original commit? 

But I am curious about this virtual directory you talked about, and how that moves the .git folder but not the project folder... somehow?

AArnott:
And by the way, I generally like to add unit tests with each of these features (in this case, to the build integration tests file). Can this virtual directory thing be emulated in a test?

Issue #96, opened 2016-12-26
Feature request: Allow "major" as one of the values for assemblyVersion.precision


asherber:
I'd like to be able to have my `AssemblyVersion` only pick up the major part of the `AssemblyFileVersion`. (Why? See https://codingforsmarties.wordpress.com/2016/01/21/how-to-version-assemblies-destined-for-nuget/)

I think this might be as simple as adding `Major` to the `VersionPrecision` enum and then tweaking `VersionOracle.GetAssemblyVersion` (and adding major to `version.schema.json`).

AArnott:
Agreed. I have no objection to enabling this. I may not get to it in the next few days so if you're interested (as you seem to have already reviewed the necessary changes) and want to send a PR, I'd be happy to review and accept it.

AArnott:
BTW, for what it's worth, I believe major.minor is a better assembly version than just major. The reason is I'd rather see an exception stating it couldn't find version 3.4 of my assembly than an exception saying it couldn't find a particular method. Since semver would suggest (or at least the way I interpret it) that new API warrants a bump to the minor version number, any new features will require a newer assembly version, it guarantees a useful exception. And binding redirects (which NuGet and MSBuild both automatically generate as necessary nowadays) take care of making sure that if I compiled against 3.2, it will still run against 3.3 (but not 3.1) which is exactly the behavior I want. But if I only can compile against 3.0 (because minor version is always zero) then there's no documentation or runtime enforcement that I actually get at least version 3.2.

asherber:
Thanks much for the response. I don't disagree with your analysis – I'm still finding my way around what works best – but I have had good results with the method outlined in that article, and I think it's worthwhile to have that option.

PR has been submitted.

PR #97, opened 2016-12-26
Allow "major" as a value for assemblyVersion.precision

asherber:
Addresses #96. In addition to the items listed there, I also had to replace two instances of `default(VersionPrecision)` with `VersionPrecision.Minor`.

### DIFF
--- a/src/NerdBank.GitVersioning/AssemblyVersionOptionsConverter.cs
+++ b/src/NerdBank.GitVersioning/AssemblyVersionOptionsConverter.cs
@@ -45,7 +45,7 @@
             var data = value as VersionOptions.AssemblyVersionOptions;
             if (data != null)
             {
-                if (data.Precision == default(VersionOptions.VersionPrecision))
+                if (data.Precision == VersionOptions.VersionPrecision.Minor)
                 {
                     serializer.Serialize(writer, data.Version);
                     return;
--- a/src/NerdBank.GitVersioning/VersionOptions.cs
+++ b/src/NerdBank.GitVersioning/VersionOptions.cs
@@ -162,7 +162,7 @@
             /// </summary>
             /// <param name="version">The assembly version (with major.minor components).</param>
             /// <param name="precision">The additional version precision to add toward matching the AssemblyFileVersion.</param>
-            public AssemblyVersionOptions(Version version, VersionPrecision precision = default(VersionPrecision))
+            public AssemblyVersionOptions(Version version, VersionPrecision precision = VersionPrecision.Minor)
             {
                 this.Version = version;
                 this.Precision = precision;
@@ -368,6 +368,11 @@
         /// </summary>
         public enum VersionPrecision
         {
+            /// <summary>
+            /// The first integer is the last number set. The rest will be zeros.
+            /// </summary>
+            Major,
+
             /// <summary>
             /// The second integer is the last number set. The rest will be zeros.
             /// </summary>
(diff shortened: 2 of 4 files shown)


asherber:
My bad, will fix. I'm having some problems getting your toolchain installed per instructions.

asherber:
Ugh. The problem is that inserting `Major` as the first member of the enum, where it belongs, screws up all kinds of default handling (since `Major` is now the default instead of `Minor`). 

I can fix this with a bunch of small changes, or I can fix it by making `Major` the last member in the enum (not intuitive, but less intrusive). What do you think?

asherber:
Sorry, I'm not having success with this -- I'm getting errors installing your toolchain, which means I can't build and run tests locally, which makes it hard to see what I'm missing. I'll close this pull request for now and come back to it when I can

AArnott:
I don't think anything needs to be installed besides vs2015 for your work. What are you struggling with? 

Regarding Major, I suspect the right thing is to insert it at top, and fix up defaults.

asherber:
`init.ps1` complains that it can't find the specified version of `NuProj`, and `build.ps1` throws a whole bunch of compilation errors in the `Tests` project. (Type or namespace does not exist, type or namespace does not exit.)

In VS 2015 Pro Update 3, I don't have the node.js or NuProj tools installed, so those projects don't load, but I see the same compilation errors in the Tests project

AArnott:
The errors in build are probably because `init.ps1` doesn't finish. Can you please copy in the exact error output of that script?

asherber:
```
PS> .\init.ps1
Restoring NuGet packages...
Unable to find version '0.10.30-beta-gcbd7693f91' of package 'NuProj'.
Restoring NPM packages...
Restoring Typings...
typings WARN deprecated 4/28/2016: "registry:dt/camel-case#0.0.0+20160316155526" is deprecated (updated, replaced or rem
oved)
typings WARN deprecated 11/21/2016: "registry:dt/node#6.0.0+20160915134512" is deprecated (updated, replaced or removed)


+-- del
| `-- pinkie-promise
|   `-- es6-promise
+-- request@2.69.0
| +-- form-data
| `-- tough-cookie
+-- camel-case (global)
`-- node (global)

P:\visual studio\nerdbank.gitversioning
PS>
```

AArnott:
Thanks. That should be working. The package it can't find is in one of the nuget sources: https://www.myget.org/feed/aarnott/package/nuget/NuProj/0.10.30-beta-gcbd7693f91

Very strange. I'll look into this. Can you file a new issue with this init.ps1 output to track?

asherber:
Now that I'm past that, I'm getting this from both `build.ps1` and VS:
```
BuildIntegrationTests.cs(854,107): error CS0122: 'ThisAssembly' is inaccessible due to its protection level [P:\visual studio\nerdbank.gitversioning\src\NerdBank.GitVersioning.Tests\NerdBank.GitVersioning.Tests.csproj]
```

asherber:
It looks like no `xxx.Version.cs` file is being generated for the `Tests` project, so it's trying to read the (internal) class from the main project.

AArnott:
It seems like you may be using a version of NuGet.exe other than the one init.ps1 downloads. 
How did you get past the NuProj restore failure?

asherber:
Thanks for the NuGet fix. I'm now able to build and debug, and I see where tests are failing.

In `VersionOracle.GetAssemblyVersion()`, can you tell me why you set `assemblyVersion` to one value, and then immediately set it to a different value? It looks to me like that first line can just be ignored. I need to tweak this function, and I want to make sure I understand what you've got there. 
It looks to me like that first line can just be ignored.

PR #100, opened 2016-12-27
Allow "major" as a value for assemblyVersion.precision

asherber:
All tests passing now. Thanks for all of your assistance in getting the environment working.


asherber:
(Fixes #97)

PR #101, opened 2016-12-27
Explicitly set the build platform, suppress loading of PS profile.

jazzdelightsme:
Also add a -MsBuildVerbosity param to build.ps1.

Some people default to using a platform-specific msbuild window, but our
project only supports 'AnyCPU'.

-NoProfile will help PS to load faster, and also prevent problems when
people have weird stuff in their $profile.


AArnott:
These look good. Thank you.

PR #104, opened 2016-12-30
Don't fall over if there is no git repo and AssemblyVersion field is specified in version.json

### LINKED ISSUE #103: What should the behavior be when there is no repo?
Sometimes I'm a lurker, so rather than fork/clone a repo, I just click the "clone or download" button and choose to just download a .zip of a project.

When NBGV is used with such a project (such as NBGV itself), it doesn't work (build fails when at some point it tries to put together a System.Version object with a negative build component).

There's a tiny tweak that can be made to make it work... but maybe we want to do more than that: maybe we want to be able to handle this scenario, so we add a proper test for it; or maybe we want to add a better error that clearly says "hey there, you can't build this project if you don't have a .git...". Or maybe we want to somehow put the "there was no repo when this code was built" into the version information embedded in the binaries.

(I haven't really looked at the test code so I don't know but I suspect it could be a pain/significant change to simulate not having a repo.)

jazzdelightsme:
Fixes #103.

### DIFF
--- a/src/NerdBank.GitVersioning/VersionOracle.cs
+++ b/src/NerdBank.GitVersioning/VersionOracle.cs
@@ -278,6 +278,9 @@
 
         private static Version GetAssemblyVersion(Version version, VersionOptions versionOptions)
         {
+            // If there is no repo, "version" could have uninitialized components (-1).
+            version = version.EnsureNonNegativeComponents();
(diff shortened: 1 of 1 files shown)


AArnott:
Thank you for the fix.

PR #107, opened 2017-02-10
Handle .git files, for submodules

### LINKED ISSUE #106: Git height not calculated on built submodules
I'm still trying to work out what's going on, and perhaps this use case is out of scope. But I wanted to report it here, for starters.

I have a couple of projects in their own dirs under a solution dir. The solution dir also has a folder called submodules, which contains a couple of git submodules. ProjA has a dependency on subProjB and subProjC, and all 3 use GitVersioning:
```
solution-dir
├───projA      
├───submodules         
│   ├───subProjB                      
│   ├───subProjC                   
```
When I build projA, the version info in the assembly is correct. But the version info in the subProjB and subProjC assemblies which get built and copied into the projA bin dir is not correct – they correctly pick up the base info from their own version.json, but the git height is 0 and the commit hash is the **hash of projA**.

The same thing happens if I open up `submodules/subProjB` directly in VS and build it, although of course it works fine if subProjB is off in its own folder somewhere else.

I can investigate this further when I have a few minutes.

asherber:
Fixes #106 

Not ready for review yet; I need appveyor to build so I can try it in the real world.


asherber:
I'm not familiar enough with libgit2sharp to construct a repo with submodules for unit tests, but I can confirm that this works as expected in the wild. The submodule projects pick up the correct version height regardless of how they are built.

asherber:
It's a little unorthodox, because the method is private and static, but I added some tests for `FindGitDir()` to show that it behaves as expected. (I could do this just with the file system, not needing to actually create a repo with submodules.)

AArnott:
Thanks for the test. I don't like to test non-public methods directly because it causes tests to be brittle in the face of refactoring of internals and by then I've forgotten what a test was good for. I'll try to whip up a real public-facing test tonight.

asherber:
Yeah, I agree with you about the tests; I was just trying to do what I can. There's also a school of thought that says if you find yourself wanting to test a private method, it should probably be a public method in a different class .....

Regarding your refactoring in 18c29, are you not worried about `File.ReadLines()` possibly throwing an exception because it turns out to be some odd binary file, or because of permissions? Maybe I was being overly cautious there.

AArnott:
> There's also a school of thought that says if you find yourself wanting to test a private method, it should probably be a public method in a different class .....

Yup, and I follow that, when I want to be in the business of supporting that particular method being called by others rather than being free to delete it when I don't like it any more. In this case, I consider it an implementation detail rather than a feature of a versioning library to be able to find submodule boundaries.

> are you not worried about File.ReadLines() possibly throwing an exception because it turns out to be some odd binary file, or because of permissions?

I think failing for a legitimate reason (an unexpected comment as the first line of the .git file, for example) is more likely than finding a binary file called `.git`. And if such a failure occurred, I'd want an error to show up rather than to swallow it and make it super hard to figure out. 
As for permissions issues, if you don't have permission to read that file, git would probably fail too. So I think I'm ok with that until a real scenario is reported as suffering from it.

asherber:
Okay, thanks, sounds good. I appreciate your looking at this so quickly!

AArnott:
I wouldn't have spent much time on it for a while if you hadn't solved the problem for me. :) Thanks.

AArnott:
libgit2sharp evidently doesn't support creating submodules yet. So I settled for checking in a 10KB binary file that contains a repro git repo.

PR #111, opened 2017-02-18
Add support for MSBuild Core (aka `dotnet build`)

### LINKED ISSUE #56: Support for MSBuild Core
Target CoreCLR with the MSBuild task and reference [MSBuild Core assemblies](https://dotnet.myget.org/gallery/dotnet-buildtools).

Remaining work items:

* [x] libgit2sharp needs to be a bit smarter about adding the native binaries directory to the PATH than just to use `AppContext.BaseDirectory` because that doesn't work for an MSBuild Task that uses it.
* [x] figure out how to automate the build of a NuPkg with all the right files in place.
* [x] Load the right MSBuild Tasks assembly based on whether MSBuild is running on .NET Fx or .NET Core
* [x] Re-implement `ThisAssembly` code generation with Roslyn APIs
* [x] Get tests running again (on both net45 and .NET Core)
* [x] Get this building on appveyor
* [x] Check that NPM package contains a nbgv.nuget folder that contains a subset of what is in the nupkg rather than a bunch of facade assemblies under build\msbuildcore

To work on Linux:

* [ ] ~~Fix native libgit2 library loading in libgit2sharp~~ (https://github.com/AArnott/Nerdbank.GitVersioning/issues/112)

AArnott:
This migrates the library projects to the new .NET SDK (aka .NET Core) project type of VS2017 and NB.GV now works in `dotnet build`. 

We *should* be very close to this working on linux/OSX as well, although we have problems with finding the native binaries on those systems.

Closes #56


el2iot2:
Am I correct, given this PR, that [this doc]
(https://github.com/AArnott/Nerdbank.GitVersioning/blob/master/doc/dotnet-cli.md) is now out of date? There it says "Sorry. DNX and dotnet CLI don't support extensible versioning systems." Can I use NB.GV with a `dotnet build` vs2017+csproj setup?

AArnott:
Quite right, @el2iot2. Thank you for calling that out. I have [updated the doc](079d24cbca3bf0872c9a0c951920dfe0498519e6)

PR #120, opened 2017-04-20
Use xUnit 2.2 and set debug type to full to get Test Explorer to work

onovotny:
DebugType needs to be full for .NET Desktop tests or they show up as `External`


AArnott:
Thanks, @onovotny. I accidentally pushed a merge onto your topic branch, so I merged your original commit and a commit that fixed appveyor's test run into master.

PR #132, opened 2017-06-07
Add option to choose SemVer option

### LINKED ISSUE #126: Switch NuGetPackageVersion to be based on SemVer2 when the nuget gallery is ready
When NuGet/NuGetGallery#3547 is fixed, flip the `NuGetPackageVersion` property to be set to `SemVer2` instead of `SemVer1`. 
Either that or have some easy way to update it using an MSBuild prop. Goal is to make it easy to switch once NuGet supports it.

As [suggested by @onovotny](https://github.com/AArnott/Nerdbank.GitVersioning/pull/119#issuecomment-304738925).

onovotny:
Closes #126 by allowing version file to specify semver level to use for package versions

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -1,6 +1,0 @@
-<Project>
-  <ItemGroup>
-    <ProjectCapability Remove="DependenciesTree" />
-    <ProjectCapability Remove="LaunchProfiles" />
-  </ItemGroup>
-</Project>
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -1,0 +1,6 @@
+<Project>
+  <ItemGroup>
+    <ProjectCapability Remove="DependenciesTree" />
+    <ProjectCapability Remove="LaunchProfiles" />
+  </ItemGroup>
+</Project>
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -72,6 +72,8 @@ public class BuildIntegrationTests : RepoTestBase
         }
     }
 
+    private string CommitIdShort => this.Repo.Head.Commits.First().Id.Sha.Substring(0, 10);
+
     protected override void Dispose(bool disposing)
     {
         Environment.SetEnvironmentVariable("_NBGV_UnitTest", string.Empty);
@@ -279,7 +281,7 @@ public class BuildIntegrationTests : RepoTestBase
         this.AssertStandardProperties(VersionOptions.FromVersion(new Version(majorMinorVersion)), buildResult);
 
         Version version = this.Repo.Head.Commits.First().GetIdAsVersion();
-        Assert.Equal($"{version.Major}.{version.Minor}.{buildResult.GitVersionHeight}", buildResult.NuGetPackageVersion);
+        Assert.Equal($"{version.Major}.{version.Minor}.{buildResult.GitVersionHeight}+g{this.CommitIdShort}", buildResult.NuGetPackageVersion);
     }
 
     [Fact]
@@ -828,7 +830,7 @@ public class BuildIntegrationTests : RepoTestBase
     {
         int versionHeight = this.Repo.GetVersionHeight(relativeProjectDirectory);
         Version idAsVersion = this.Repo.GetIdAsVersion(relativeProjectDirectory);
-        string commitIdShort = this.Repo.Head.Commits.First().Id.Sha.Substring(0, 10);
+        string commitIdShort = this.CommitIdShort;
         Version version = this.Repo.GetIdAsVersion(relativeProjectDirectory);
         Version assemblyVersion = GetExpectedAssemblyVersion(versionOptions, version);
         var additionalBuildMetadata = from item in buildResult.BuildResult.ProjectStateAfterBuild.GetItems("BuildMetadata")
@@ -859,10 +861,11 @@ public class BuildIntegrationTests : RepoTestBase
         Assert.Equal(expectedBuildMetadata, buildResult.SemVerBuildSuffix);
 
         // NuGet is now SemVer 2.0 and will pass additional build metadata if provided
+        bool semVer1 = versionOptions?.NuGetPackageVersion?.SemVer == 1;
         string pkgVersionSuffix = buildResult.PublicRelease
-            ? $"{(additionalBuildMetadata.Any() ? "+" : string.Empty)}{string.Join(".", additionalBuildMetadata)}"
-            : expectedBuildMetadata;
-        Assert.Equal($"{idAsVersion.Major}.{idAsVersion.Minor}.{idAsVersion.Build}{versionOptions.Version.Prerelease}{pkgVersionSuffix}", buildResult.NuGetPackageVersion);
+            ? (semVer1 ? string.Empty : expectedBuildMetadata)
+            : (semVer1 ? $"-g{commitIdShort}" : $"-g{commitIdShort}{(additionalBuildMetadata.Any() ? "+" : string.Empty)}{string.Join(".", additionalBuildMetadata)}");
+        Assert.Equal($"{idAsVersion.Major}.{idAsVersion.Minor}.{idAsVersion.Build}{GetSemVerAppropriatePrereleaseTag(versionOptions)}{pkgVersionSuffix}", buildResult.NuGetPackageVersion);
 
         var buildNumberOptions = versionOptions.CloudBuild?.BuildNumber ?? new VersionOptions.CloudBuildNumberOptions();
         if (buildNumberOptions.Enabled)
@@ -894,9 +897,11 @@ public class BuildIntegrationTests : RepoTestBase
         }
     }
 
-    private static string GetSemVer1PrereleaseTag(VersionOptions versionOptions)
+    private static string GetSemVerAppropriatePrereleaseTag(VersionOptions versionOptions)
     {
-        return versionOptions.Version.Prerelease?.Replace('.', '-');
+        return versionOptions.NuGetPackageVersion?.SemVer == 1
+            ? versionOptions.Version.Prerelease?.Replace('.', '-')
+            : versionOptions.Version.Prerelease;
     }
 
     private async Task<BuildResults> BuildAsync(string target = Targets.GetBuildVersion, LoggerVerbosity logVerbosity = LoggerVerbosity.Detailed, bool assertSuccessfulBuild = true)
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -281,7 +281,7 @@ public class BuildIntegrationTests : RepoTestBase
         this.AssertStandardProperties(VersionOptions.FromVersion(new Version(majorMinorVersion)), buildResult);
 
         Version version = this.Repo.Head.Commits.First().GetIdAsVersion();
-        Assert.Equal($"{version.Major}.{version.Minor}.{buildResult.GitVersionHeight}+g{this.CommitIdShort}", buildResult.NuGetPackageVersion);
+        Assert.Equal($"{version.Major}.{version.Minor}.{buildResult.GitVersionHeight}", buildResult.NuGetPackageVersion);
     }
 
     [Fact]
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -861,7 +861,7 @@ public class BuildIntegrationTests : RepoTestBase
         Assert.Equal(expectedBuildMetadata, buildResult.SemVerBuildSuffix);
 
         // NuGet is now SemVer 2.0 and will pass additional build metadata if provided
-        bool semVer1 = versionOptions?.NuGetPackageVersion?.SemVer == 1;
+        bool semVer1 = (versionOptions?.NuGetPackageVersion ?? VersionOptions.NuGetPackageVersionOptions.DefaultInstance).SemVer == 1;
         string pkgVersionSuffix = buildResult.PublicRelease
             ? (semVer1 ? string.Empty : expectedBuildMetadata)
             : (semVer1 ? $"-g{commitIdShort}" : $"-g{commitIdShort}{(additionalBuildMetadata.Any() ? "+" : string.Empty)}{string.Join(".", additionalBuildMetadata)}");
@@ -899,7 +899,7 @@ public class BuildIntegrationTests : RepoTestBase
 
     private static string GetSemVerAppropriatePrereleaseTag(VersionOptions versionOptions)
     {
-        return versionOptions.NuGetPackageVersion?.SemVer == 1
+        return (versionOptions.NuGetPackageVersion ?? VersionOptions.NuGetPackageVersionOptions.DefaultInstance).SemVer == 1
             ? versionOptions.Version.Prerelease?.Replace('.', '-')
             : versionOptions.Version.Prerelease;
     }
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -580,11 +580,15 @@ public class BuildIntegrationTests : RepoTestBase
 
     [Theory]
     [PairwiseData]
-    public async Task BuildNumber_VariousOptions(bool isPublic, VersionOptions.CloudBuildNumberCommitWhere where, VersionOptions.CloudBuildNumberCommitWhen when, [CombinatorialValues(0, 1, 2)] int extraBuildMetadataCount)
+    public async Task BuildNumber_VariousOptions(bool isPublic, VersionOptions.CloudBuildNumberCommitWhere where, VersionOptions.CloudBuildNumberCommitWhen when, [CombinatorialValues(0, 1, 2)] int extraBuildMetadataCount, [CombinatorialValues(1, 2)] int semVer)
     {
         var versionOptions = BuildNumberVersionOptionsBasis;
         versionOptions.CloudBuild.BuildNumber.IncludeCommitId.Where = where;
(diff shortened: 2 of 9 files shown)


src/NerdBank.GitVersioning/version.schema.json
```diff
@@ -39,6 +39,12 @@
       "minimum": 1,
       "maximum": 6
     },
+    "packageSemVerLevel": {
+      "type": "string",
```
AArnott: Also, I suspect it's not appropriate to use just one field to choose this. There are, after all, two packages types (NPM and NuGet) that might need independent values.

src/NerdBank.GitVersioning.Tests/VersionOracleTests.cs
```diff
@@ -149,4 +149,22 @@ public void SemVer1PrereleaseConversionPadding()
...
     }
+
+    [Fact]
+    public void SemVer2ForPackageOption()
+    {
+        VersionOptions workingCopyVersion = new VersionOptions
+        {
+            Version = SemanticVersion.Parse("7.8.9-foo.25"),
+            BuildNumberOffset = 2,
+            SemVer1NumericIdentifierPadding = 3,
```
AArnott: Why bother setting this semver1 setting and the `BuildNumberOffset` if you don't use it?

onovotny: Will fix, just copied the earlier test.

src/NerdBank.GitVersioning/version.schema.json
onovotny: I can split those, no reason not to. As to Number, I though an enum spelling it out made more sense than a 1 or a 2 but I can do either. Just lmk.

onovotny: Another thought -- what about `packageVersionStyle` as the property, so others could be added later?
`npmPackageVersionStyle` and `nugetPackageVersionStyle`?

AArnott: We already have `assemblyVersion` as a precedent where it lets a particular rendition of a version be customized, either by setting an explicit version or customizing how it is created.
So perhaps we should follow a similar schema for `packageVersion`:

```json
"nugetPackageVersion": {
  "semVer": 1
},
"npmPackageVersion": {
  "semVer": 2
}
```

Of course we can hold off defining NPM unless you feel like it. I don't know if it supports 1 or 2 or both and if there's any reason to not just support the latest for them.

onovotny: Seems like the course of "YAGNI" would say to just add `nugetPackageVersion` and leave NPM alone. That said, what should it default to, semver 1 or semver 2? I suspect 2?

AArnott: Agreed: let's not define the npm one unless we know it's helpful. From the NB.GV code it looks like we use semver1 so I'm *guessing* NPM doesn't support semver2. But it's a moot point if we aren't going to define the `npmPackageVersion` field at this stage.

As for NuGet, I think I'd prefer we default to semver2. NB.GV's master branch is a 2.0-beta after all, so it's a good time to set defaults that are current for the times.

onovotny: Okay, I'll update the PR to only define `nugetPackageVersion` with a `semVer` number property with a default of 2, and leave the NPM as SemVer1.

Will update later this eve.

src/NerdBank.GitVersioning/version.schema.json
```diff
@@ -39,6 +39,12 @@
...
+      "default": "semVer1",
+      "description": "The SemVer level to use for package version strings.",
+      "enum": [ "semVer1", "semVer2" ]
```
AArnott: Can we use `range` or some other schema field to force the value to be 1 or 2?

onovotny: Yeah, in-line with the above, I'll add the right schema field.

onovotny:
Okay, I think it's ready....  the submodule tests seem to be sporadically failing on AppVeyor due to weirdness: https://ci.appveyor.com/project/AArnott/nerdbank-gitversioning/build/2.0.10-beta+g4283c03ae9#L160

src/NerdBank.GitVersioning/version.schema.json
onovotny: done

src/NerdBank.GitVersioning/version.schema.json
onovotny: Done

onovotny:
Also, here are the relevant specs:
https://github.com/NuGet/Home/wiki/SemVer-2.0.0-support#spec
https://github.com/NuGet/Home/wiki/SemVer2-support-for-nuget.org-%28server-side%29#identifying-semver-v200-packages

NuGet allows the build metadata but does normalize it such that it's not part of the package identity. Thus, I think NB.GV should pass-through whatever the user specifies that's SemVer2 and let NuGet sort it out...?

Means that `1.0.0-preview.1+A0` is allowed, but it'll be seen as `1.0.0-preview.1`. `3.2.1+fixes` is also allowed and seen as `3.2.1`. That means you cannot also upload `3.2.1+more.fixes` since the identity would clash.

src/Directory.Build.targets
```diff
@@ -0,0 +1,6 @@
+<Project>
+  <ItemGroup>
+    <ProjectCapability Remove="DependenciesTree" />
+    <ProjectCapability Remove="LaunchProfiles" />
```
AArnott: What are these for, and why are they both in props and targets?

AArnott:
@onovotny  any concerns about my changes before I merge?

onovotny:
The props/targets changes should be reverted, I accidentally checked them in. It's related to troubleshooting perf issues in 15.3.

src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
```diff
@@ -859,10 +861,11 @@ private void AssertStandardProperties(VersionOptions versionOptions, BuildResult
         Assert.Equal(expectedBuildMetadata, buildResult.SemVerBuildSuffix);
 
         // NuGet is now SemVer 2.0 and will pass additional build metadata if provided
+        bool semVer1 = versionOptions?.NuGetPackageVersion?.SemVer == 1;
         string pkgVersionSuffix = buildResult.PublicRelease
-            ? $"{(additionalBuildMetadata.Any() ? "+" : string.Empty)}{string.Join(".", additionalBuildMetadata)}"
-            : expectedBuildMetadata;
-        Assert.Equal($"{idAsVersion.Major}.{idAsVersion.Minor}.{idAsVersion.Build}{versionOptions.Version.Prerelease}{pkgVersionSuffix}", buildResult.NuGetPackageVersion);
+            ? (semVer1 ? string.Empty : expectedBuildMetadata)
+            : (semVer1 ? $"-g{commitIdShort}" : $"-g{commitIdShort}{(additionalBuildMetadata.Any() ? "+" : string.Empty)}{string.Join(".", additionalBuildMetadata)}");
```
onovotny: Not a huge deal, but the statement on L866-867 seems really hard to understand as a series of string interpolated ternaries.

onovotny:
One thought...while SemVer2 does allow the buildmetadata to be tucked into the release packages, not sure if that'll filter it out of the clients or it gets normalized away completely. Either way, if there's a user-visible bit, it may be worth another option within `nugetPackageVersion` to disable build metadata on public packages. I can add that if you want... what should the default be?

onovotny:
Based on this, https://github.com/NuGet/Home/wiki/SemVer2-support-for-nuget.org-%28server-side%29#identifying-semver-v200-packages, it seems like packages with just buildmetadata and no prerelease packages are still marked as SemVer2 only and will be filtered from older clients. I haven't tested/confirmed that though.

If that's the case, it may make sense to hold off on putting build metadata on release packages by default for now since that'd probably have too much of an adverse effect. I'm happy to add an option for it though as per above, with the default value you think is best.

onovotny:
Had a third idea, which I'm implementing now... 

`"includeBuildMetaData": "prerelease" // optional. Include the build metadata in SemVer 2 NuGet versions. Values are 'prerelease', 'all', or 'none' Default is 'prerelease'. `

This would enable the user to select if they want build metadata attached to all, prerelease or no package versions. Default would be prerelease only so release packages work with all clients.

AArnott:
> .while SemVer2 does allow the buildmetadata to be tucked into the release packages, not sure if that'll filter it out of the clients or it gets normalized away completely

The way they explained it to me, the build metadata gets added to the nuspec as a tag and that's it. But it's worth testing. 

> If that's the case, it may make sense to hold off on putting build metadata on release packages by default for now since that'd probably have too much of an adverse effect.

This only happens if the user wants semver2. If they want semver1, there's already a switch for that. I'm not sure why we need a second switch?

> This would enable the user to select if they want build metadata attached to all, prerelease or no package versions. Default would be prerelease so release packages work with all clients.

I don't think prerelease is the right switch here. If you're going to maximize reach when it matters, I'd change that to "publicRelease".

onovotny:
I tested it and it's a problem. It's not about public release vs non-public because my RTW version can't be 1.0.0+foobar. That will now limit people who consumes my package to NuGet 4.3+.

AArnott:
> I tested it and it's a problem.

I presume the same problem applies to using `.` delimiters in the prerelease portion. So it sounds like folks who don't mind filtering packages to NuGet 4.3 can use semver 2 and those concerned can use semver 1. The default value for this is still an interesting question. Do you agree?

onovotny:
My thoughts are that it's okay to limit prerelease users to SemVer 2 / NuGet 4.3+ but release packages shouldn't have that limit.

That's why I was trying to go with a "use semver 2 for prerelease packages" and SemVer1 / no build-metadata for stable packages. This is different than public/non-public because I will publish the prerelease packages on nuget.org and that's a "public release"

AArnott:
OK. I didn't understand that you were intentionally limiting scope on your public prerelease packages. 

For the sake of limiting the number of cross cutting options for ease of use (I personally was having trouble figuring how your proposed additional property would affect other properties in various configurations), I'd rather just have the single semver switch you added. If you want prereleases to follow semver2 rules, then you can certainly set it to `2` at the same time you add `-beta` to the version.json file, then switch it back to `1` when you trim off the `-beta` prerelease. Sure, it's two things to change instead of one, but I think it's really odd if you have semver set to 2 but have another property that basically says "don't really use 2--at least not in this particular way".

onovotny:
I see what you're saying, but my concern is that it may make it more complicated to "do the right thing" for a majority of users. Assuming users want to use SemVer 2 by default, and I'd wager that most do, they shouldn't have to know that they need to use SemVer 1 for a stable NuGet release or risk compatibility.

AArnott:
If folks are concerned about compatibility with existing nuget clients, they should use semver1. I don't know why someone who wants a broader audience at RTW releases would want to shrink their audience at a prerelease artificially. 

Given only NuGet 4.3+ supports semver2 and that nuget.org itself recommends 4.1 instead of what is really the latest, perhaps we should leave the default at semver 1. Does that alleviate the concern?

onovotny:
I'm indifferent about the defaults between SemVer 1 and SemVer 2. I can see arguments either way and the config makes it easy to decide.

What I'm concerned about is that people won't realize that adding the BuildMetadata to the stable version will cause issues.

> I don't know why someone who wants a broader audience at RTW releases would want to shrink their audience at a prerelease artificially.

The reason is simple -- as a project maintainer, I don't care as much if you can't use my packages in prerelease if you're using older tools. I'm going to do what's easier for me and use SemVer 2 for its better sorting and cleaner package names. When it's time for public release though, I don't want to exclude those folks.

Here's another reason -- aesthetics and opinion. Just because you can add build metadata to a NuGet package doesn't necessarily mean everyone wants to. I'm not convinced I do. I want SemVer 2, but I'm not sold on being required to put the git commit id in it. In my opinion, it makes the package version too ugly and long to read. Others may have different opinions, so I wanted to provide a choice.

AArnott:
> What I'm concerned about is that people won't realize that adding the BuildMetadata to the stable version will cause issues.

What issues would that be? Build metadata is dropped in semver1. And in Semver2, it doesn't because the user has already indicated they are willing to take the semver2 hit.

> but I'm not sold on being required to put the git commit id in it. In my opinion, it makes the package version too ugly and long to read. Others may have different opinions, so I wanted to provide a choice.

I would agree with you... but I thought the build metadata was stripped from the package version and only put into the nuspec file. Is that not the case? If build metadata shows up with the version on the gallery, then I agree that's a problem and I'll very willingly stop tacking that on for public releases.

onovotny:
@AArnott It shows up in the public gallery. It's in the nuspec version element. https://www.myget.org/gallery/zeroconf

AArnott:
Ok, ya that doesn't look great. Do we know how it appears (or will appear) on nuget.org?

onovotny:
One way to find out...pushing it now :)

AArnott:
That's the way to do it: 

image

As this is so new, I wonder if myget.org would be willing to follow nuget.org's lead on this.

AArnott:
What do you think? Would this be an acceptable presentation?

onovotny:
Check your email ... it's not the same on both galleries. I wonder what their plan is.
https://preview.nuget.org/packages/Zeroconf/2.9.0-preview.4
https://www.nuget.org/packages/Zeroconf/2.9.0-preview.4

onovotny:
Either way, I'm less concerned now but it depends on what they say the behavior is client-wise for a stable release.... because it still is in the nuspec file and older clients could choke on it..

AArnott:
Interesting. I've never seen preview.nuget.org. It looks like they hide build metadata altogether. I don't mind that either.
The fact that it shows up on myget.org seems to be OK even if they don't fix it, assuming you use myget as a CI or prerelease system, as that aligns with your goal of smaller audience and potentially less convenient. Is that correct?

> but it depends on what they say the behavior is client-wise for a stable release.... because it still is in the nuspec file and older clients could choke on it..

Can we test that? Say, push a _stable_ package with build metadata to nuget.org then use the 3.3 client to try to install it?

onovotny:
Yeah, will test that now...

AArnott:
Thanks. Alternatively an unstable package is fine too, but one without a `.` delimiter in the prerelease portion.

onovotny:
Email is more current

AArnott:
For now I've changed semver default back to 1. It seems premature that we would require 4.3 by default before the nuget team themselves were willing to recommend anything beyond 4.1.

The question as to what we do in the semver2 case is of course still open.

onovotny:
Given what we know now, I'm against putting built metadata in the NuGet package versions considering their UI hides it. I agree it should go somewhere, just not there. I'd like to use SemVer2 for the prerelease portions of the tag though. Happy to make whatever updates/configuration you think is needed to support that.

AArnott:
Sounds reasonable. If nuget ends up with a good story for build metadata, or 4.3 becomes predominant, we can revisit. Let's:

1. stop appending *any* build metadata, regardless of prerelease or publicRelease. 
1. we should continue to append the git commit ID as a prerelease tag for all non-public releases.
1. leave semver 1 as default.

Does that sound about right?

onovotny:
Sounds good. I assume these changes should just be for the `NuGetPackageVersion` flag and the `SemVer2` property should be as-is? I don't see any reason not to leave the SemVer2 property as it is since it is correct.

So to recap, for a SemVer 1 nonpublic, it'd be

`1.0.0-beta-001-g<hash>`

and SemVer2 non-public
`1.0.0-beta.1-g<hash>`? Does that even work? or should SemVer2 non-public include it as metadata `1.0.0-beta.1+g<hash>`

Public releases would not have the metadata for either.

AArnott:
> I assume these changes should just be for the NuGetPackageVersion property and the SemVer2 property should be as-is? I don't see any reason not to leave the SemVer2 property as it is since it is correct.

Hmmm... while that's true, I think I'd rather keep `NuGetPackageVersion` to be exactly equivalent to either semver1 or semver2 rather than it being yet something else. 

>  SemVer2 non-public `1.0.0-beta.1-g<hash>`

In Semver2, identifiers are delimited by `.` not `-`. So the commit ID should be preceded by a period, otherwise the `beta.1` won't be sorted properly. So I think it should be: `1.0.0-beta.1.g<hash>`

And I should clarify: if build metadata is specified in version.json itself, I think we should retain that in semver2. It's just the build metadata that nb.gv *adds* for git commit ID that should never be in the `+buildMetadata` portion.

> I think the commit id has to be build metadata for the non-public semver 2 case?

No it doesn't. For non-public packages, they are *always* prerelease packages because we add `-g<hash>` to the end of the version.

onovotny:
Does it allow more than four parts? I didn't think `1.0.0-beta.1.g<hash>` was legal... I'm ok with that if is.

AArnott:
> Does it allow more than five parts? 

Does *what* allow more than five parts? I am unaware of any limitation to the number of identifiers in a prerelease tag.

onovotny:
Argh, you're right, mis-parsed that as being part of the version and not pre-release tag. Just pushed an update that should implement this for your review.

AArnott:
With the changes I pushed on top, I'm ready to complete the PR. Waiting for final appveyor validation. I'd appreciate your final sign off as well.

onovotny:
Will check it out once there's a successful build. Any idea why that one test is intermittently failing?

AArnott:
I believe the intermittent test failure is a bug in the 7zip nuget package the test uses to unpack. I haven't taken time yet to investigate it, but it's getting increasingly annoying.

onovotny:
Any reason it can't use the built in compression libraries?

AArnott:
> Any reason it can't use the built in compression libraries?

They're dog slow. But reliability trumps performance. So that's probably the way to go.

AArnott:
I've filed a bug against 7z.NET: https://github.com/patrickb8man/7z.NET/issues/2

onovotny:
Quick smoke test works as expected with Zeroconf

AArnott:
Thanks for your contribution.

onovotny:
Any chance of getting an updated version on NuGet soonish? :)

AArnott:
publishing now.

onovotny:
Thanks!

Issue #114, opened 2017-03-14
Build perf is negatively impacted


AArnott:
Running PerfView during the build of the PInvoke project, I see that almost a full second goes to the GetBuildVersion task, which seems to be doing some repeat work perhaps within the same task execution, and certainly across projects. Can we cache the results across projects?

```
Name                                                                                                                                                            	Inc %	   Inc
 nerdbank.gitversioning.tasks!GetBuildVersion.Execute                                                                                                           	  2.2	   939
+ nerdbank.gitversioning!VersionOracle.Create                                                                                                                   	  2.2	   935
|+ nerdbank.gitversioning!Nerdbank.GitVersioning.VersionOracle..ctor(class System.String,class LibGit2Sharp.Repository,class Nerdbank.GitVersioning.ICloudBuild)	  2.2	   919
||+ nerdbank.gitversioning!GitExtensions.GetVersionHeight                                                                                                       	  0.9	   381
||+ nerdbank.gitversioning!GitExtensions.GetIdAsVersion                                                                                                         	  0.8	   340
||+ nerdbank.gitversioning!VersionFile.GetVersion                                                                                                               	  0.4	   181
```

Opportunities:

* [ ] Add an incremental build aspect to the `GetBuildVersion` target where the results are saved to a file with the cache key being the tuple: version.json timestamp (and path?) and commit id. (@japj's [idea](https://github.com/dotnet/Nerdbank.GitVersioning/issues/114#issuecomment-669697460))
* [ ] Cache the height for one or more commits so even if the commit advances, we quickly find a commit whose height was predetermined in the cache file and cut off the potentially long git history traversal (@djluck's [idea](https://github.com/dotnet/Nerdbank.GitVersioning/issues/114#issuecomment-669730257)) 
* [x] #508 Cut down on redundant `GetBuildVersion` executions per-project that @djluck [identified](https://github.com/dotnet/Nerdbank.GitVersioning/issues/114#issuecomment-669713622), possibly by using the `MSBuild` task call to invoke another target but with a reduced set of global properties to improve target skipping around P2P and TargetFramework invocations.
* [x] #508 Build on the shared MSBuild target invocation within a single project to share it across projects based on a project property that specifies the directory path for which version.json will be searched for and used as the 'cache key' for reuse. This key can be a global property in the MSBuild task invocation to automatically cache at the right level, and this same property can serve as the input for the version.json search.
* [x] #507 Never parse the same version.json content twice)
[...]

Issue #94, opened 2016-12-20
Support for native binaries?


jazzdelightsme:
I've got a .sln with both native (C/C++) and managed (C# and C++/CLI) projects. It would be great to also be able to stamp the native binaries with the same version information as the managed binaries. I'm not sure what the best way to implement this would be... perhaps to generate a .rc file, or to produce the .res file directly.

AArnott:
Ooh,  that's interesting. I'd be very interested in a pull request on this if someone can do it.

heaths:
I've got a pattern I can generalize and work on a code change here.

AArnott:
Thanks, @heaths

jazzdelightsme:
Cool!

I happen to have some self-contained C# code that can emit a .res file containing version information (VS_FIXEDFILEINFO et al). I've used it in other, managed code projects where you couldn't depend on having rc.exe to compile a .rc to a .res. For an actual native code project, you _could_ depend on being able to compile an .rc, but if I were to try to do this change, I don't know that it would be any less annoying to codegen a .rc file than it would to just create the .res file, given that I know how and already have code to do the latter. Also, since link.exe can accept multiple .res files, it's not a composition problem like it is for managed code (with csc.exe, you get one /win32res:file).

But after snooping around this project (NBGV) a bit, I've come to the conclusion that my msbuild-fu is far too weak, and given the amount of vacation time I have left, I don't think I'd be able to get very far. But @heaths, you are welcome to the .res-generating code if it is useful at all.

If generating the .res directly is interesting, another thought I had is maybe the .res-emitting code could/should be its own nuget package, since it's potentially useful in other projects. I could probably scrape that together.

Code here: https://gist.github.com/jazzdelightsme/844a5e6b97a58148767e01eb6dac5c9b

AArnott:
Personally, I favor NB.GV generating source code rather than a binary file so that folks can see what effect it's having on the build more clearly.

heaths:
My plan is to generate source, yes. It can be combined with other resources in the RC file and be used as variables in code.

- Heath via Nine<http://www.9folders.com/> on Android
________________________________

AArnott:
Sweet!

heaths:
Seems we shouldn't close this until I merge my script changes with @robmen's changes to just support this. Right now, manual edits will be necessary unlike _csproj_ or _vbproj_ files.

AArnott:
[...]

Issue #125, opened 2017-05-29
Unstable test: Submodule_RecognizedWithCorrectVersion


AArnott:
The `VersionOracleTests.Submodule_RecognizedWithCorrectVersion` test frequently fails with [this output](https://ci.appveyor.com/project/AArnott/nerdbank-gitversioning/build/2.0.4-beta+g4a5336233c#L719):

```
    VersionOracleTests.Submodule_RecognizedWithCorrectVersion [FAIL]
      System.ComponentModel.Win32Exception : The system cannot find the file specified
      Stack Trace:
           at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
           at System.Diagnostics.Process.Start()
           at SevenZipNET.WProcess.Execute()
           at SevenZipNET.SevenZipExtractor.ExtractAll(String destination, Boolean overwrite, Boolean keepStructure)
        src\NerdBank.GitVersioning.Tests\TestUtilities.cs(70,0): at Nerdbank.GitVersioning.Tests.TestUtilities.ExtractRepoArchive(String repoArchiveName)
        src\NerdBank.GitVersioning.Tests\VersionOracleTests.cs(18,0): at VersionOracleTests.Submodule_RecognizedWithCorrectVersion()
```

It causes PR builds to fail frequently.

onovotny:
Seems like this can be closed now with your recent commit.

AArnott:
Closed by e1be2046c9680029c68b974c357e635e847a5dd8

AArnott:
It's still happening.

PR #134, opened 2017-06-11
Partial performance improvement for #114

robmen:
I was starting to work towards #94 but hit some performance issues and found open issue #114. I took a couple hours to understand the code then took a shot at minimizing the number of git calls (I think I've reduced the 5 or 6 calls to 1 or 2).

I didn't delete as much code from `GitExtensions.cs` as I'd like to since there were a lot of unit tests using the code. If this PR looks good, I can also reduce the code in `GitExtensions.cs` (and `GitExtensionsTests.cs`) to clean up the bit of duplication between `VersionOracle.cs` and `GitExtensions.cs` remaining.

Note, there are a couple code clean up commits in the PR. The performance improvement commit is in the middle. I tried to be pretty detailed in the commit message on that commit.

### DIFF
--- a/src/NerdBank.GitVersioning/VersionOracle.cs
+++ b/src/NerdBank.GitVersioning/VersionOracle.cs
@@ -5,9 +5,7 @@
     using System.Globalization;
     using System.IO;
     using System.Linq;
-    using System.Text;
     using System.Text.RegularExpressions;
-    using System.Threading.Tasks;
     using Validation;
 
     /// <summary>
@@ -21,9 +19,9 @@
         private static readonly Regex NumericIdentifierRegex = new Regex(@"(?<![\w-])(\d+)(?![\w-])");
 
         /// <summary>
-        /// The cloud build suppport, if any.
+        /// The 0.0 version.
         /// </summary>
-        private readonly ICloudBuild cloudBuild;
+        private static readonly Version Version0 = new Version(0, 0);
 
         /// <summary>
         /// Initializes a new instance of the <see cref="VersionOracle"/> class.
@@ -51,24 +49,33 @@
         /// </summary>
         public VersionOracle(string projectDirectory, LibGit2Sharp.Repository repo, ICloudBuild cloudBuild)
         {
-            this.cloudBuild = cloudBuild;
-            this.VersionOptions =
-                VersionFile.GetVersion(repo, projectDirectory) ??
-                VersionFile.GetVersion(projectDirectory);
-
             var repoRoot = repo?.Info?.WorkingDirectory?.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
             var relativeRepoProjectDirectory = !string.IsNullOrWhiteSpace(repoRoot)
                 ? projectDirectory.Substring(repoRoot.Length).TrimStart(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar)
                 : null;
 
             var commit = repo?.Head.Commits.FirstOrDefault();
+
+            var committedVersion = VersionFile.GetVersion(commit, relativeRepoProjectDirectory);
+
+            var workingVersion = VersionFile.GetVersion(projectDirectory);
+
+            this.VersionOptions = committedVersion ?? workingVersion;
+
             this.GitCommitId = commit?.Id.Sha ?? cloudBuild?.GitCommitId ?? null;
-            this.VersionHeight = repo?.GetVersionHeight(relativeRepoProjectDirectory) ?? 0;
+            this.VersionHeight = CalculateVersionHeight(relativeRepoProjectDirectory, commit, committedVersion, workingVersion);
             this.BuildingRef = cloudBuild?.BuildingTag ?? cloudBuild?.BuildingBranch ?? repo?.Head.CanonicalName;
 
             // Override the typedVersion with the special build number and revision components, when available.
-            this.Version = repo?.GetIdAsVersion(relativeRepoProjectDirectory, this.VersionHeight) ?? this.VersionOptions?.Version.Version;
-            this.Version = this.Version ?? new Version(0, 0);
+            if (repo != null)
+            {
+                this.Version = GetIdAsVersion(commit, committedVersion, workingVersion, this.VersionHeight);
+            }
+            else
+            {
+                this.Version = this.VersionOptions?.Version.Version ?? Version0;
+            }
+
             this.VersionHeightOffset = this.VersionOptions?.BuildNumberOffset ?? 0;
 
             this.PrereleaseVersion = ReplaceMacros(this.VersionOptions?.Version.Prerelease ?? string.Empty);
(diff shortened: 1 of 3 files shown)


AArnott:
Thanks, @robmen! I look forward to reviewing this... probably tomorrow. 
Did you happen to measure the perf impact of the change? Was the perf on your repo bad before? If so, do you attribute it to the number of projects building or the number of commits that are counted when calculating the git version height?

robmen:
I didn't measure exactly (it was late) but each "call" to git to find the version file would end up walking the git tree to find the matching version. It isn't *just* calculating the height. There was lots of duplicate work when I was stepping through debugger. A fake repository I created with a hundred commits was brutal to walk all the way down and back up several times.

I found it because the `VersionOptions` were recreated and compared with the working tree in many places. I was experimenting with a feature to override the BuildNumberOffset passed in via MSBuild property (conversation for a different pull request). I stopped counting after finding 5 different instances `VersionOptions` (all read from the same `version.json`) whereI had to overwrite the offset to make it "stick". That's how I ended up walking all the way down the tree in the debugger and why I created a repo with a bunch of commits to see if I was seeing things correctly and walking down even deeper loops in the debugger and then finding performance issue #114 on GitHub and finally saying to myself, "Oh, this is doing a lot more work than necessary" (and this is a crazy run on sentence).

I believe this change also has the possibility to reduce a lot of code. I left it all in because there were so many unit tests using the code. I didn't know if you wanted the code factored that way (for example, if the NB.GV assembly was being used for other things). Basically everything is `public` so hard to know the desired surface area. :smile:

AArnott:
> Basically everything is public so hard to know the desired surface area. 

I had this idea that the library that the MSBuild task calls into would be itself useful. So ya, a lot is public. 
But what really *needs* to be public is whatever the MSBuild task, the two ps1 scripts, and the NPM module needs to function.

robmen:
That's exactly what it looked like. And the "API" provided by that NB.GV assembly is pretty friendly for one-off requests but if you need to do several operations (like the MSBuild task does) you end up with lots of repeat (and potentially costly) git work. As you'll see my change reduces calls to the NB.GV assembly.

If you don't want the public surface going forward, I can reduce the overall code count quite a bit. But that's up to you. So far, it sounds like I guessed correctly. 😉

AArnott:
This PR has test failures that should be dealt with. One of them is #125 and can be ignored. But there is another test failure that I haven't seen before on this PR build.

src/NerdBank.GitVersioning.Tests/GitExtensionsTests.cs
```diff
@@ -123,22 +123,6 @@ public void GetVersionHeight_VersionJsonHasParsingErrorsInHistory()
     }
 
     [Fact]
-    public void GetTruncatedCommitIdAsInteger_Roundtrip()
```
robmen: No loss of coverage. These test were covering code that was deleted.

src/NerdBank.GitVersioning/VersionFile.cs
```diff
@@ -34,15 +32,14 @@ public static class VersionFile
         /// <param name="commit">The commit to read the version file from.</param>
         /// <param name="repoRelativeProjectDirectory">The directory to consider when searching for the version.txt file.</param>
         /// <returns>The version information read from the file.</returns>
-        public static VersionOptions GetVersion(LibGit2Sharp.Commit commit, string repoRelativeProjectDirectory = null)
+        public static VersionOptions GetVersionFromCommit(LibGit2Sharp.Commit commit, string repoRelativeProjectDirectory = null)
```
robmen: No. The code was harder to understand when trying to tease out number of times the repository was being accessed. Now that I've straightened out the code, I'll try reverting these changes back to the overloads.

robmen:
I do not understand why these failures are happening on the build machine but not locally. I'll make the changes above and try to reproduce the issue.

robmen:
Actually, I can reproduce with all the latest changes all integrated. Investigating now.

src/NerdBank.GitVersioning/VersionFile.cs
robmen: I'm putting the code back but in doing so, I remember why I had such a hard time with the overloads. The behavior based on the input parameters is very different. Getting the version information from a commit is very different behavior from getting a version by searching the directory tree. Thus, the different name for the methods was *extremely* helpful for me.

But I'm reverting that change now and several others to simplify the salient part of the fix. Sorry, I shouldn't have included all the extra change originally (but the changes helped me learn the code).

robmen:
Okay, I simplified this PR down to just the optimization (which is the most important part). It should be much easier to see changes now. Good call.

Also, I now understand how `GitExtensions.cs` has an API surface that is larger than what NB.GV needs and that explains a lot of the "dead code" I was tripping across.

src/NerdBank.GitVersioning/VersionOracle.cs
```diff
@@ -391,5 +398,41 @@ private static string MakePrereleaseSemVer1Compliant(string prerelease, int padd
...
+        }
+
+        private static bool IsVersionFileChangedInWorkingTree(VersionOptions committedVersion, VersionOptions workingVersion)
+        {
+            if (workingVersion != null)
+            {
+                return !EqualityComparer<VersionOptions>.Default.Equals(workingVersion, committedVersion);
+            }
+
+            return false; // a missing working version is allowed and not a change.
```
AArnott: Does this mean that if I delete my `version.json` file, I'll still get the same version and height calculation? Why is that? In the version of this method that is in `GitExtensions.cs`, we consider a missing version file in the working copy to be a change. We only return false from that method if it's a "bare" repo, which doesn't seem to be considered here. Was that intentional?

AArnott:
Thanks, @robmen. This iteration looks much easier to assess. 🙂

src/NerdBank.GitVersioning/VersionOracle.cs
robmen: No. I was trying to match the logic you describe. But none of the tests covered the case and this ended up giving the right answer within the set of things that I understood.  For example, I removed the bare repo check because it failed tests (but everything passed without it). However, I will quickly admit that I may have missed a use case that doesn't have a test. :smile:

Let me play with the code with this new information and see if I get the code to match your explanation and pass the tests.

AArnott: I may be missing a test for bare repos. I can't think of anyone that actually needs this and I can't remember why I supported it to begin with. So I'll accept this PR as-is, or whenever you're satisfied. Just let me know.

robmen: Heh, at this point, you know better than I. This is working for all my test cases right now and what is in the unit tests. If an issue comes up in the future, a test could be created for the test case and everything fixed... then we'd know for sure.

Really your call.

robmen: Okay, I had some time and just stared at the code a little longer. I think I understand the deleted work tree case better and I think the bare repo case devolves to it. I'll have one last change right now.

AArnott:
Thanks again.

PR #135, opened 2017-06-15
Support GitVersioning in .vcxprojs

### LINKED ISSUE #94: Support for native binaries?

robmen:
Generate a .rc file and include it into the build pipeline for native
code in a fashion similar to the `GenerateAssemblyVersionInfo` task.

Fixes #94


robmen:
Build failure is related to deprecated NPM module, not my code:

> .\node_modules\.bin\typings : typings WARN deprecated 4/27/2016: "registry:dt/camel-case#0.0.0+20160316155526" is deprecated (updated, replaced or removed)
> At C:\projects\nerdbank-gitversioning\init.ps1:33 char:13
> +             .\node_modules\.bin\typings install
> +             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>     + CategoryInfo          : NotSpecified: (typings WARN de...ced or removed):String) [], RemoteException
>     + FullyQualifiedErrorId : NativeCommandError
> 
> typings WARN deprecated 11/21/2016: "registry:dt/node#6.0.0+20160915134512" is deprecated (updated, replaced or removed

heaths:
Recommend also pulling in the migration/initialization scripts from https://github.com/AArnott/Nerdbank.GitVersioning/compare/master...heaths:issue94#diff-82f8d5aaf8a67007cec2e056c725cfea, or I can do so later and test.

src/Nerdbank.GitVersioning.Tasks/NativeVersionInfo.cs
```diff
@@ -0,0 +1,307 @@
...
+
+    public class NativeVersionInfo : Task
+    {
+        private const int VFT_APP = 0x1;
+        private const int VFT_DLL = 0x2;
+
+        private const string FileHeaderComment = @"------------------------------------------------------------------------------
+ <auto-generated>
+     This code was generated by a tool.
+     Runtime Version:4.0.30319.42000
```
heaths: May as well populate the runtime version with the actual version at runtime.

src/Nerdbank.GitVersioning.Tasks/NativeVersionInfo.cs
```diff
@@ -0,0 +1,307 @@
...
+            internal void AddDefine(string name, int value)
+            {
+                this.codeBuilder.AppendLine($"#define {name} {value}");
+            }
+
+            internal void AddDefine(string name, string value)
+            {
+                var escapedValue = "\"" + value.Replace("\\", "\\\\") + "\"";
+
+                this.codeBuilder.AppendLine($"#define {name} {escapedValue}");
```
heaths: Should define ASCII and Unicode variants like https://github.com/heaths/Nerdbank.GitVersioning/blob/e451d79b1d04b76b249579f8827953294de809cc/src/Nerdbank.GitVersioning.Tasks/VersionHeader.cs#L43-L62. Seems like this would be a good place to do it.

src/Nerdbank.GitVersioning.Tasks/NativeVersionInfo.cs
robmen: Meh. This header matches `AssemblyVersionInfo.cs` (actually tried to share the code but decided it made things more complicated). I'll leave as is for consistency and both could be fixed in future.

src/Nerdbank.GitVersioning.Tasks/NativeVersionInfo.cs
robmen: Very good idea. Fixed.

robmen:
Also, I'll leave migration/initialization scripts to others. I don't actually use that functionality (actively avoid it) so I'm not a good candidate to improve them. Looks like you already have a huge head start on that work.

heaths:
@AArnott I can submit a PR for the scripts after you merge this if you'll hold out on a release (I'll watch for the merge notification and submit a PR based on @robmen's branch).

AArnott:
> Build failure is related to deprecated NPM module

No, it's your code. Don't let AppVeyor's red highlighting distract you: it's just a warning.
The actual break is due to the failure of test `BuildIntegrationTests.NativeVersionInfo_CreateNativeResourceDll` later down the log. And that looks very related to your PR.

src/Nerdbank.GitVersioning.NuGet/build/Nerdbank.GitVersioning.targets
```diff
@@ -2,7 +2,11 @@
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
   <PropertyGroup>
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
-    <VersionSourceFile>$(IntermediateOutputPath)\$(AssemblyName).Version$(DefaultLanguageSourceExtension)</VersionSourceFile>
```
AArnott: I'm curious why this had to be moved. I can't think of anyone who depends on this property being defined in MSBuild's evaluation model, but it's a risk. So is it justified?

src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
```diff
@@ -802,6 +802,29 @@ public static object[][] BuildNumberData
...
+
+        string targetFile = Path.Combine(this.projectDirectory, result.BuildResult.ProjectStateAfterBuild.GetPropertyValue("TargetPath"));
+        Assert.True(File.Exists(targetFile));
+
+        var fileInfo = FileVersionInfo.GetVersionInfo(targetFile);
+        Assert.Equal("1.2", fileInfo.FileVersion);
+        Assert.Equal("1.2.0", fileInfo.ProductVersion);
+        Assert.Equal("test", fileInfo.InternalName);
+        Assert.Equal("NerdBank", fileInfo.CompanyName);
+        Assert.Equal("Copyright (c) 2017. All rights reserved.", fileInfo.LegalCopyright);
```
AArnott: Ticking time bomb here would fail in 2018. I'll update to use the year at the time the test runs to match the product behavior.

AArnott:
Thanks.

src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
```diff
@@ -822,7 +822,7 @@ public static object[][] BuildNumberData
         Assert.Equal("1.2.0", fileInfo.ProductVersion);
         Assert.Equal("test", fileInfo.InternalName);
         Assert.Equal("NerdBank", fileInfo.CompanyName);
-        Assert.Equal("Copyright (c) 2017. All rights reserved.", fileInfo.LegalCopyright);
+        Assert.Equal($"Copyright (c) {DateTime.Now.Year}. All rights reserved.", fileInfo.LegalCopyright);
```
robmen: Heh. Thanks!

src/Nerdbank.GitVersioning.NuGet/build/Nerdbank.GitVersioning.targets
robmen: I moved it because the value is now different for managed vs. native code. The `$(DefaultLanguageSourceExtension)` for C++ code is `.cpp` but we're generating a `.rc` file so I localized and specialized the value in the targets that used them.

If you're worried about the breaking change I could keep this value up here and override it in the native code target. Or could just use a different variable local to the target or global here (although I don't know what I'd call it 😄 ).

Do you want me to do one of those?

AArnott: At this point, no. Thanks for the background of this change. I'm ok to wait to hear that it broke someone before changing it. But now that I know the reasoning, we can change it with confidence should the need arise.

robmen:
For my education, what was the unit test failure that broke AppVeyor? I ran all the tests locally and had them pass so I'm curious what I missed.

AArnott:
The failure was here : https://ci.appveyor.com/project/AArnott/nerdbank-gitversioning/build/2.0.25-beta+ga6ab42abce#L164

But I've never seen it before or since. I didn't fix it either. It just passed on rerun. :frown:

robmen:
Hmm. That's a pattern I reused from the tests in the same file. The `NerdBank.GitVersioning.targets` are extracted at test time. I'm not able to reproduce it locally... timing issue?

Something to keep an eye out.

PR #137, opened 2017-06-19
Allow BuildNumberOffset to be overridden via MSBuild

### LINKED ISSUE #136: Allow BuildNumberOffset to be overridden
Today the build number offset is specified in the version file. In migrating to NB.GV, we've hit a complex scenario where part of our build needs one build number offset and in another pass the same projects need to be built with a different build number offset.

This feature request is to allow the build number offset to be overridden via a property in MSBuild.

The alternative we can think of right now is to modify the source version file during the build process. We definitely don't want to do that as it creates ordered dependencies in the build (that would majorly disrupt parallel build) and we'd have to clean the file up after every build.

robmen:
Fixes #136

### DIFF
--- a/src/NerdBank.GitVersioning/VersionOracle.cs
+++ b/src/NerdBank.GitVersioning/VersionOracle.cs
@@ -29,8 +29,9 @@
         /// <param name="projectDirectory"></param>
         /// <param name="gitRepoDirectory"></param>
         /// <param name="cloudBuild"></param>
+        /// <param name="overrideBuildNumberOffset"></param>
         /// <returns></returns>
-        public static VersionOracle Create(string projectDirectory, string gitRepoDirectory = null, ICloudBuild cloudBuild = null)
+        public static VersionOracle Create(string projectDirectory, string gitRepoDirectory = null, ICloudBuild cloudBuild = null, int? overrideBuildNumberOffset = null)
         {
             Requires.NotNull(projectDirectory, nameof(projectDirectory));
             if (string.IsNullOrEmpty(gitRepoDirectory))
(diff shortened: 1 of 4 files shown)


src/NerdBank.GitVersioning/VersionOracle.cs
```diff
@@ -60,6 +61,19 @@ public VersionOracle(string projectDirectory, LibGit2Sharp.Repository repo, IClo
 
             var workingVersion = VersionFile.GetVersion(projectDirectory);
 
+            if (overrideBuildNumberOffset.HasValue)
+            {
+                if (committedVersion != null)
+                {
+                    committedVersion.BuildNumberOffset = overrideBuildNumberOffset.Value;
```
AArnott: Are you sure overwriting the original `BuildNumberOffset` value is what you want and need?
*Modifying* the offset by a given value sounds like it might be more generally useful. For example, if `version.json` set the offset at `5`, and you wanted to shift everything about 10000, then would 10005 be better than 10000?

robmen: That is a totally reasonable option as well. I approached the problem as a way to "override" the value independent of `version.json`. But it could be an "adjustment" instead.

The "override" (current) implementation matches my need directly but the "adjustment" method could be made to work just as well. I mildly prefer the former but if you prefer the latter I can adjust the code (and unittest).

AArnott: Well, as you are the only customer for this feature (and I hope it remains that way) I'm inclined to go with what you prefer. I just wanted to make sure you'd considered the alternative.

src/NerdBank.GitVersioning/VersionOracle.cs
```diff
@@ -29,8 +29,9 @@ public class VersionOracle
         /// <param name="projectDirectory"></param>
         /// <param name="gitRepoDirectory"></param>
         /// <param name="cloudBuild"></param>
+        /// <param name="overrideBuildNumberOffset"></param>
         /// <returns></returns>
-        public static VersionOracle Create(string projectDirectory, string gitRepoDirectory = null, ICloudBuild cloudBuild = null)
+        public static VersionOracle Create(string projectDirectory, string gitRepoDirectory = null, ICloudBuild cloudBuild = null, int? overrideBuildNumberOffset = null)
```
AArnott: This is a breaking public API change and I think will break the Get-Version.ps1 script (which regrettably we have no automated tests for yet). 
Can you please update that script to accommodate the change and test that the script works?

Thanks.

robmen: I noticed that and tested the `Get-Version.ps1` script. :smile:

I considered creating a new `Create` methods with this parameter but the default parameter worked so I didn't. Want me to create an overload instead?

src/NerdBank.GitVersioning/VersionOracle.cs
robmen: Yeah. I don't like playing games with version numbers either. But this was the most robust solution of several bad solutions. And by far the most straight forward to test.

src/NerdBank.GitVersioning/VersionOracle.cs
AArnott: Thanks. I'm impressed: I didn't think powershell honored default parameters.

PR #138, opened 2017-06-21
Added CloudBuild support for Jenkins

khellang:
Because Jenkins doesn't have built-in support for setting the build name/number, I decided to just do everything; write the number to a file, echo it to standard out (so you could use RegEx to pick it up) and set the `GIT_VERSION` environment variable. The user can then pick the method they want to set the build number. I've tested this on our project and it works well :smile:


src/NerdBank.GitVersioning/CloudBuild.cs
```diff
@@ -17,6 +17,7 @@ public static class CloudBuild
             new VisualStudioTeamServices(),
             new TeamCity(),
             new AtlassianBamboo(), 
+            new Jenkins()
```
AArnott: nit: can you add the trailing comma to this line so that the next addition doesn't have to touch your line?

src/NerdBank.GitVersioning/CloudBuildServices/Jenkins.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
```
AArnott: stylecop: please put `using` directives inside the namespace block.

src/NerdBank.GitVersioning/CloudBuildServices/Jenkins.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public string BuildingTag => null;
+
+        public bool IsPullRequest => false;
+
+        public string BuildingBranch => CloudBuild.ShouldStartWith(Environment.GetEnvironmentVariable("GIT_BRANCH"), "refs/heads/");
+
+        public string GitCommitId => Environment.GetEnvironmentVariable("GIT_COMMIT");
+
+        public bool IsApplicable => !string.IsNullOrEmpty(this.GitCommitId);
```
AArnott: I could imagine other CI servers defining a `GIT_COMMIT` environment variable. Can you find another one to test for that is more specific to Jenkins? It looks like `JENKINS_URL` might be a good choice.

src/NerdBank.GitVersioning/CloudBuildServices/Jenkins.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                { name, value }
+            };
+        }
+
+        private static void WriteVersionFile(string buildNumber)
+        {
+            var workspacePath = Environment.GetEnvironmentVariable("WORKSPACE");
+
+            var versionFilePath = Path.Combine(workspacePath, "version.txt");
```
AArnott: `version.txt` is actually a file that NB.GV looks for (it was a feature before version.json came along). Can we instead name this file `jenkins_build_number.txt`? Or is `version.txt` somehow magic to Jenkins and not something we can control? I don't see any reference to `version.txt` anywhere else in this code so I'm curious who reads it and how they find it.

src/NerdBank.GitVersioning/CloudBuildServices/Jenkins.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// </remarks>
+    internal class Jenkins : ICloudBuild
+    {
+        private static readonly Encoding UTF8NoBOM = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+
+        public string BuildingTag => null;
+
+        public bool IsPullRequest => false;
+
+        public string BuildingBranch => CloudBuild.ShouldStartWith(Environment.GetEnvironmentVariable("GIT_BRANCH"), "refs/heads/");
```
AArnott: I see GitVersion uses this expression for the environment variable:
`(Environment.GetEnvironmentVariable("GIT_LOCAL_BRANCH") ?? Environment.GetEnvironmentVariable("GIT_BRANCH"))` which suggests to me that perhaps either environment variable may be defined.

src/NerdBank.GitVersioning/CloudBuildServices/Jenkins.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public string GitCommitId => Environment.GetEnvironmentVariable("GIT_COMMIT");
+
+        public bool IsApplicable => !string.IsNullOrEmpty(this.GitCommitId);
+
+        public IReadOnlyDictionary<string, string> SetCloudBuildNumber(string buildNumber, TextWriter stdout, TextWriter stderr)
+        {
+            WriteVersionFile(buildNumber);
+
+            stdout.WriteLine($"## GIT_VERSION: {buildNumber}");
```
AArnott: What is it that does the magic? Writing the file or emitting this string?

src/NerdBank.GitVersioning/CloudBuildServices/Jenkins.cs
khellang: So `GIT_LOCAL_BRANCH` is configurable (to which local branch do you want to check out the changes) and doesn't contain the remote. I don't think that's what we want...

src/NerdBank.GitVersioning/CloudBuildServices/Jenkins.cs
khellang: Yep :+1:

src/NerdBank.GitVersioning/CloudBuildServices/Jenkins.cs
khellang: None. Or all. As I mentioned, there's really no magic in Jenkins that does this. You need a plugin, like "Build Name Setter", which can be configured to use either option. Whatever the user wants. Since I couldn't see a way to configure specific cloud builds, I'm not sure we should make that choice for the user, so I went with all :stuck_out_tongue_closed_eyes:

src/NerdBank.GitVersioning/CloudBuildServices/Jenkins.cs
khellang: Sure, this is just an arbitrary name (mentioned as an example in the plugin docs). This can be whatever if the current name is a problem.

src/NerdBank.GitVersioning/CloudBuildServices/Jenkins.cs
AArnott: This property shouldn't return any information about the remote. It should just be a local branch (like 'master' or 'refs/heads/master')

AArnott:
Thanks!

Issue #37, opened 2015-12-19
Build Server support: set build server's build number to git version height during build


onovotny:
Might be useful to have a way to set the build number the build server uses to match the calculated version.

I realize that all build servers have different ways to do this but it is possible:
- TeamCity
- AppVeyor
- VSTS Agents

To name a few popular ones

AArnott:
Are you suggesting that we make it so TeamCity is influenced by NBGV, or the other way around?

Build server build numbers tend to be ever-incrementing integers. But NBGV build numbers are calculated based on git commit height. So they are intended to represent different things. I think build servers determine the build number to use before they even start the build, so it's far too late for NBGV to influence what that build number is. 

Most build server integration I see in packages like these let the influence go the other way: the build number determined by the build server becomes the build number used in actual version stamping of the binaries. But that's exactly why I wrote this package: I don't believe in that. I think builds should be exactly reproducible by syncing to the commit that was originally built. The build server's stateful incrementing integer shouldn't make break that.

AArnott:
I sympathize though. I have often wish the numbers aligned. But then I think about how sometimes I need to build the same commit more than once, and I realize the build server's numbering scheme makes sense to keep the 2 builds separate.

onovotny:
No, I meant that NBGV would set the build number...each build server has a way of having its version set programmatically and is implemented here by GitVersion:
https://github.com/GitTools/GitVersion/tree/master/src/GitVersionCore/BuildServers

AArnott:
Oh! That's *very* interesting. Thanks for the tip.

AArnott:
The work for VSTS is done via #61 and commit 923fccde2ca19f88a913808308ace119ff83b3a7

AArnott:
Hmmm... well, I guess I don't go as far as updating the build number itself -- I set other VSTS variables though.

AArnott:
Working on AppVeyor cloud build number support now. [Online documentation](https://www.appveyor.com/docs/api/projects-builds#update-project-build-number) suggests it's simpler than making HTTP calls the way GitVersion does.

KirillOsenkov:
Hey @AArnott there's something I don't understand about how this works with AppVeyor. If you look at the history here:
https://ci.appveyor.com/project/KirillOsenkov/msbuildstructuredlog/history
[...]

PR #139, opened 2017-06-22
Implements CloudBuild support for Teamcity

Romanx:
Hi there,

We're using this project to version with teamcity and noticed that the cloud build support wasn't implemented.

I took @onovotny's advice and made the implementation very similar to GitVersion. I tried adding tests for the integration, hopefully i've done all that needs to be done.

When setting up with teamcity you have to manually expose the branch name through a environment variable, I called this `BUILD_GIT_BRANCH` however the name can easily be changed. I'll update the cloud build docs to include this.

#37

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -436,6 +436,8 @@ public class BuildIntegrationTests : RepoTestBase
             new object[] { CloudBuild.AppVeyor.SetItem("APPVEYOR_REPO_BRANCH", branchName) },
             new object[] { CloudBuild.VSTS.SetItem("BUILD_SOURCEBRANCH", $"refs/heads/{branchName}") },
             new object[] { CloudBuild.VSTS.SetItem("BUILD_SOURCEBRANCH", branchName) }, // VSTS building a github repo
+            new object[] { CloudBuild.Teamcity.SetItem("BUILD_GIT_BRANCH", $"refs/heads/{branchName}") },
+            new object[] { CloudBuild.Teamcity.SetItem("BUILD_GIT_BRANCH", branchName) },
         };
     }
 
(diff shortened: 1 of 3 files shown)


src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
```diff
@@ -436,6 +436,8 @@ public static IEnumerable<object[]> CloudBuildOfBranch(string branchName)
             new object[] { CloudBuild.AppVeyor.SetItem("APPVEYOR_REPO_BRANCH", branchName) },
             new object[] { CloudBuild.VSTS.SetItem("BUILD_SOURCEBRANCH", $"refs/heads/{branchName}") },
             new object[] { CloudBuild.VSTS.SetItem("BUILD_SOURCEBRANCH", branchName) }, // VSTS building a github repo
+            new object[] { CloudBuild.Teamcity.SetItem("BUILD_GIT_BRANCH", $"refs/heads/{branchName}") },
+            new object[] { CloudBuild.Teamcity.SetItem("BUILD_GIT_BRANCH", branchName) },
```
AArnott: Does TeamCity really need to double up like VSTS does? Or is a single entry sufficient as it is for AppVeyor? It's really based on whether TeamCity reliably specifies a fully-qualified branch name or does not, or if it interchangeably uses both syntaxes. I suspect it reliably does just one.

src/NerdBank.GitVersioning/CloudBuildServices/TeamCity.cs
```diff
@@ -18,11 +18,15 @@ internal class TeamCity : ICloudBuild
...
         public IReadOnlyDictionary<string, string>  SetCloudBuildNumber(string buildNumber, TextWriter stdout, TextWriter stderr)
         {
+            (stdout ?? Console.Out).WriteLine($"##teamcity[buildNumber '{buildNumber}']");
             return new Dictionary<string, string>();
         }
 
         public IReadOnlyDictionary<string, string> SetCloudBuildVariable(string name, string value, TextWriter stdout, TextWriter stderr)
         {
+            (stdout ?? Console.Out).WriteLine($"##teamcity[setParameter name='GitVersion.{name}' value='{value}']");
+            (stdout ?? Console.Out).WriteLine($"##teamcity[setParameter name='system.GitVersion.{name}' value='{value}']");
```
AArnott: These cloud build variables are not always about versions. Is the `GitVersion.` prefix that is always added really appropriate?
The purpose of this method is to expose several build properties to the build server for use in later build steps. And they typically already contain sufficient distinguishing names as to infer whether they are version related.

src/NerdBank.GitVersioning/CloudBuildServices/TeamCity.cs
```diff
@@ -6,7 +6,7 @@
```
AArnott: Can you please add a comment with a link to the docs from TeamCity you referred to in writing this?

Romanx: Added changes to the `cloudbuild.md`

Romanx:
@AArnott Changes have been made. You're right about the branch name. The way I recommend doing it always returns the `ref/heads/{branchName}` format. The prefix has been removed and my notes have been added the `cloudbuild.md` let me know if you need any more adjustments

AArnott:
Thanks!

PR #146, opened 2017-09-04
Also set GitBuildVersionSimple cloud variable

heaths:
For use cases that require just a simple semver, also set the
GitBuildVersionSimple cloud build variable from BuildVersionSimple.


AArnott:
Thanks for contributing.

PR #150, opened 2017-10-02
Add the assembly name to error messages.

bradphelan:
This makes it easier to figure out the build log especially
with parallel builds


AArnott:
I look at parallel build logs all the time. MSBuild appends the project file at the end of each error message (or at least, near enough that it's easy to find and associate using the `2>` prefix added to every single line).
Is that not enough?

PR #154, opened 2017-11-23
Add option for version.json to derive from parents folders

### LINKED ISSUE #145: Support inheritance in version.json config (or limit Update Build Number tasks)
We are using GitVersioning in a number of solutions (~66 projects total). All assemblies should be versioned the same across the codebase, so the GitVersioning package was installed in each project and we created a version.json in the repo root. We use TFS as our build server so the cloudBuild.buildNumber feature was enabled. As a result, when solutions are built, the log shows that the build number is being updated dozens of times. It seems like this is causing some of our builds to fail intermittently ("Error while copying content to a stream" or "An error occurred while sending the request").

While I'm sure the builds failing is a Microsoft issue, I'd also like to limit the number of "update build number" tasks to keep the logged output clean. One of the projects in this codebase contains core shared business logic that is referenced in every solution. I'd like to enable cloudBuild.buildNumber task in this shared project only but have it inherit the version and other settings from the root project.json file. Unfortunately, when I tried to configure this:

```
The "Nerdbank.GitVersioning.Tasks.GetBuildVersion" task failed unexpectedly.
System.NullReferenceException: Object reference not set to an instance of an object.
   at Nerdbank.GitVersioning.VersionOracle.get_PrereleaseVersion()
   at Nerdbank.GitVersioning.VersionOracle.get_AssemblyInformationalVersion()
   at Nerdbank.GitVersioning.Tasks.GetBuildVersion.ExecuteInner()
   at MSBuildExtensionTask.ContextAwareTask.Execute()
   at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
   at Microsoft.Build.BackEnd.TaskBuilder.<ExecuteInstantiatedTask>d__26.MoveNext()
```

Is there a way to inherit config or otherwise limit the number of "Update Build Number" tasks when cloudBuild output is enabled in the root across multiple projects?

AArnott:
Closes #145

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -926,7 +926,7 @@ public class BuildIntegrationTests : RepoTestBase
         Assert.Equal(expectedBuildMetadata, buildResult.SemVerBuildSuffix);
 
         // NuGet is now SemVer 2.0 and will pass additional build metadata if provided
-        bool semVer2 = (versionOptions?.NuGetPackageVersion ?? VersionOptions.NuGetPackageVersionOptions.DefaultInstance).SemVer == 2;
+        bool semVer2 = versionOptions?.NuGetPackageVersionOrDefault.SemVer == 2;
         string pkgVersionSuffix = buildResult.PublicRelease ? string.Empty : $"-g{commitIdShort}";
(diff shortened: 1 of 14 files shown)


AArnott:
To do:

* [ ] Update JSON schema
* [ ] Add test for version incrementing between inheriting files.

PR #159, opened 2017-12-18
Add option to set all cloud variables

### LINKED ISSUE #130: Add variables to cloudbuild
There may be other reasons to use cloudBuild variables in a script, like needing to manually pack NuGet satellite assemblies. 

It'd be convenient if all of the variables calculated by NV.GV could be exposed, possibly prefixed -- NBGV_{variable}

heaths:
Fixes #130 

Fixed up integration tests and added unit test. Tested in real project as well.

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -506,7 +506,7 @@ public class BuildIntegrationTests : RepoTestBase
             var versionOptions = new VersionOptions
             {
                 Version = SemanticVersion.Parse("1.0"),
-                CloudBuild = new VersionOptions.CloudBuildOptions { SetVersionVariables = true },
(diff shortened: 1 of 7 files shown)


src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
```diff
@@ -506,7 +506,7 @@ public static object[][] CloudBuildVariablesData
             var versionOptions = new VersionOptions
             {
                 Version = SemanticVersion.Parse("1.0"),
-                CloudBuild = new VersionOptions.CloudBuildOptions { SetVersionVariables = true },
+                CloudBuild = new VersionOptions.CloudBuildOptions { SetAllVariables = true, SetVersionVariables = true },
```
AArnott: How about a negative test for this as well, where this is false, and we assert that the `NBGV_` variables *aren't* set?
You could add a bool parameter to the Theory method, then assert based on that below.

AArnott: I just added that.

PR #167, opened 2018-02-06
Set Lib2Git Config SearchPaths to Empty for Global\System level. This…

benmerms:
Avoid Lib2Git to lookup and read Global\System .gitconfig files

### DIFF
--- a/src/NerdBank.GitVersioning/VersionOracle.cs
+++ b/src/NerdBank.GitVersioning/VersionOracle.cs
@@ -363,6 +363,11 @@
         {
             Requires.NotNullOrEmpty(repoRoot, nameof(repoRoot));
             var gitDir = FindGitDir(repoRoot);
+
+            // Override Config Search paths to empty path to avoid new Repository instance to lookup for Global\System .gitconfig file
+            LibGit2Sharp.GlobalSettings.SetConfigSearchPaths(LibGit2Sharp.ConfigurationLevel.Global, string.Empty);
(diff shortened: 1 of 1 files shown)


src/NerdBank.GitVersioning/VersionOracle.cs
```diff
@@ -363,6 +363,11 @@ private static LibGit2Sharp.Repository OpenGitRepo(string repoRoot)
         {
             Requires.NotNullOrEmpty(repoRoot, nameof(repoRoot));
             var gitDir = FindGitDir(repoRoot);
+
+            // Override Config Search paths to empty path to avoid new Repository instance to lookup for Global\System .gitconfig file
+            LibGit2Sharp.GlobalSettings.SetConfigSearchPaths(LibGit2Sharp.ConfigurationLevel.Global, string.Empty);
+            LibGit2Sharp.GlobalSettings.SetConfigSearchPaths(LibGit2Sharp.ConfigurationLevel.System, string.Empty);
```
AArnott: Messing with `GlobalSettings` would seem to have side effects on any other MSBuild Task running in the same process. Is there any way to set these settings for just this one repo open operation?

benmerms: I checked Repository.cs , RepositoryOptions.cs for some ways to overwrite Global Config Searchpath for 1 specific Repository instance ( or in the constructor ) .. did not find anything.

AArnott: OK. Thanks for checking.

PR #168, opened 2018-02-09
Add optional GitRelativeProjectPath parameter to GetBuildVersion for …

benmerms:
…the case where Project Path may not be in the same tree folder as Git Repo

### DIFF
--- a/src/Nerdbank.GitVersioning.Tasks/GetBuildVersion.cs
+++ b/src/Nerdbank.GitVersioning.Tasks/GetBuildVersion.cs
@@ -8,6 +8,7 @@
     using Microsoft.Build.Framework;
     using Microsoft.Build.Utilities;
     using MSBuildExtensionTask;
+    using Validation;
 
     public class GetBuildVersion : ContextAwareTask
     {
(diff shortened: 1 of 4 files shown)


src/Nerdbank.GitVersioning.Tasks/GetBuildVersion.cs
```diff
@@ -160,9 +175,19 @@ protected override bool ExecuteInner()
         {
             try
             {
+                if (!string.IsNullOrEmpty(this.ProjectPathRelativeToGitRepoRoot))
+                {
+                    Requires.Argument(Path.IsPathRooted(this.ProjectPathRelativeToGitRepoRoot), nameof(this.ProjectPathRelativeToGitRepoRoot), "Path must be relative.");
```
benmerms: if I follow your changes .... and checks ... I think it should be !Path.IsPathRooted ...  as you want the path to be relative path.

AArnott: Fixed in #169

PR #179, opened 2018-05-18
Document cloudBuild.setAllVariables field

### LINKED ISSUE #178: Docs: CloudBuild: Include all variables
Today, the cloudbuild, set variables option sets three variables. Would it make sense to set all of them and not just those three? There's various reasons they could be useful and it'd reduce the need to call the PowerShell version directly if everything was set as an env variable.

AArnott:
Fixes #178


onovotny:
Looks good. One suggestion might be to also mention that these should be called once per solution somehow?

AArnott:
Thanks. That's a good suggestion, which I'll add in another PR.

PR #180, opened 2018-05-18
Advise folks to set cloud build variables from just one project

AArnott:


AArnott:
@onovotny  What do you think?

onovotny:
That works. One other idea, that's more deterministic and perhaps a better option for this scenario, is packaging NB.GV as a .NET Core Global tool. It can return a json object like the `Get-Version.ps1` then take parameters to set the variables optionally.

This has a benefit over the `ps1` file because it's easier to invoke, especially outside powershell. Today, I do something like https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/master/Build/build.yml#L29-L32, but it'd be easier to do:

`dotnet install NB.GV`
`NB.GV /setEnvVars`

I could easily reference them everywhere I needed.

Thoughts?

AArnott:
@onovotny I hadn't heard of .NET Core Global tools. Thanks for the tip. It looks like it's new in .NET Core 2.1, so too early to document anything for it, I think. So I'll complete this PR and open an issue to track.

tzachshabtay:
Hi @AArnott it's not clear from this doc whether the "put the cloud variables in one project" advice also applies to the "buildNumber" and "setAllVariables" attributes, or just relevant to "setVersionVariables". Can you clarify?

It's also not clear if I need to set "setVersionVariables" if I already set "setAllVariables".

Thanks.

AArnott:
Thanks for the feedback, @tzachshabtay. Does #183 help?

tzachshabtay:
Thanks @AArnott, looks good.

PR #191, opened 2018-06-25
Add dotnet tool: nbgv 

### LINKED ISSUE #181: Expose as a .NET Core Global tool
As @onovotny suggested in [a PR comment](https://github.com/AArnott/Nerdbank.GitVersioning/pull/180#issuecomment-390307040), consider packaging NB.GV as a [.NET Core Global tool](https://natemcmaster.com/blog/2018/02/02/dotnet-global-tool/).

AArnott:
As a high-level summary of the new nbgv dotnet console tool, here is the top-level usage text:

```
usage: nbgv <command> [<args>]

    install        Prepares a project to have version stamps applied
                   using Nerdbank.GitVersioning.
    get-version    Gets the version information for a project.
    set-version    Updates the version stamp that is applied to a
                   project.
    tag            Creates a git tag to mark a version.
    get-commits    Gets the commit(s) that match a given version.
    cloud          Communicates with the ambient cloud build to set the
                   build number and/or other cloud build variables.
```

Additional help is available for each individual command.

Todo:

* [x] Add documentation 
* [x] Don't execute commands when `--help` trails them
* [x] get-version should accept an optional git ref for which to return version info

Closes #181

### DIFF
--- a/doc/nbgv-cli.md
+++ b/doc/nbgv-cli.md
@@ -1,0 +1,39 @@
+# Using the nbgv .NET Core CLI tool
+
+Perform a one-time install of the `nbgv` tool using the following dotnet CLI command:
+
+```cmd
+dotnet tool install -g nbgv
+```
+
+You may then use the `nbgv` tool to install Nerdbank.GitVersioning into your repos, as well as query and update version information for your repos and projects.
+
+Install Nerdbank.GitVersioning into your repo using this command from within your repo:
+
+```cmd
+nbgv install
+```
+
+This will create your initial `version.json` file.
+It will also add/modify your `Directory.Build.props` file in the root of your repo to add the `PackageReference` to the latest `Nerdbank.GitVersioning` package available on nuget.org.
+
+## Learn more
+
+There are several more sub-commands and switches to each to help you build and maintain your projects, find a commit that built a particular version later on, create tags, etc.
+
+Use the `--help` switch on the `nbgv` command or one of its sub-commands to learn about the sub-commands available and how to use them. For example, this is the basic usage help text:
+
+```cmd
+nbgv --help
+usage: nbgv <command> [<args>]
+
+    install        Prepares a project to have version stamps applied
+                   using Nerdbank.GitVersioning.
+    get-version    Gets the version information for a project.
+    set-version    Updates the version stamp that is applied to a
+                   project.
+    tag            Creates a git tag to mark a version.
+    get-commits    Gets the commit(s) that match a given version.
+    cloud          Communicates with the ambient cloud build to set the
+                   build number and/or other cloud build variables.
+```
--- a/doc/nbgv-cli.md
+++ b/doc/nbgv-cli.md
@@ -2,7 +2,7 @@
 
 Perform a one-time install of the `nbgv` tool using the following dotnet CLI command:
 
-```cmd
+```ps1
 dotnet tool install -g nbgv
 ```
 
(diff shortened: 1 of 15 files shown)


AArnott:
CC: @onovotny

src/nbgv/Program.cs
```diff
@@ -0,0 +1,430 @@
...
+            const string PackageReferenceItemType = "PackageReference";
+            const string PackageId = "Nerdbank.GitVersioning";
+            if (!propsFile.GetItemsByEvaluatedInclude(PackageId).Any(i => i.ItemType == "PackageReference"))
+            {
+                propsFile.AddItem(
+                    PackageReferenceItemType,
+                    PackageId,
+                    new Dictionary<string, string>
+                    {
+                        { "Version", "2.1.23" }, // TODO: use the latest version... somehow...
```
AArnott: I'm trying to use the NuGet.Client v4 package to do this, but [the only docs are for v3](https://docs.microsoft.com/en-us/nuget/reference/nuget-client-sdk) and v4 seems to have made it... difficult.

src/nbgv/nbgv.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <PackAsTool>True</PackAsTool>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <RootNamespace>Nerdbank.GitVersioning.Tool</RootNamespace>
+    <LangVersion>7.3</LangVersion>
+    <Description>A .NET Core Tool that can install, read and set version information based on git history, using Nerdbank.GitVersioning.</Description>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.CommandLine" Version="0.1.0-preview2-180503-2" />
```
onovotny: FWIW, I used to use this but switched to 
`Microsoft.Extensions.CommandLineUtils` https://github.com/onovotny/SignService/blob/master/src/SignClient/SignClient.csproj#L11
Since it doesn't look like the other one will ever see any updates.

Works similarly as you can see here: https://github.com/onovotny/SignService/tree/master/src/SignClient

AArnott: Interesting. I'd given Nate McMaster's fork of that library I try but was disappointed by its apparent lack of support for sub-commands. Your sample demonstrates at least that it recognizes such a thing.
I'm pretty comfortable with the System.CommandLine pattern though, and since I'm publishing the package instead of shipping it with dependenceis to nuget.org, the fact that it will never ship doesn't bother me. I've used this package several times already.
But I'll check out the MS.Ex.CLU one at some point for my next project. Thanks for the tip.

src/nbgv/Program.cs
AArnott: Thanks for the link. Interestingly enough it seems this is the line I was looking for. 

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/f0809a1fcd4fb582acd535470c30bffc69221f9e/PackageViewModel/PackageViewModel.csproj#L12

I was pulling down Nuget.Client instead of NuGet.PackageManagement.

AArnott: Hmmm.. no, that doesn't work either. And v4.7.0 doesn't even exist (on nuget.org) for that package.

onovotny: It does seem to be on Nuget.org, but they're publishing it with build-metadata, which seems confusing:

https://www.nuget.org/packages/NuGet.PackageManagement/4.7.0

doc/nbgv-cli.md
```diff
@@ -0,0 +1,39 @@
+# Using the nbgv .NET Core CLI tool
+
+Perform a one-time install of the `nbgv` tool using the following dotnet CLI command:
+
+```cmd
+dotnet tool install -g nbgv
```
onovotny: As a common use is likely to be on CI servers/build scripts to get a version, is it worth mentioning local tools?

`dotnet tool install nbvb --tool-path .`
`.\nbgv ...`

That way it won't cause issues with anything else on the machine already and doesn't mess with global state.

PR #195, opened 2018-07-16
Remove duplicate <PrivateAssets>all</PrivateAssets>

josetr:


src/Nerdbank.GitVersioning.Tasks/Nerdbank.GitVersioning.Tasks.csproj
```diff
@@ -82,9 +82,6 @@
     <PackageReference>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <ProjectReference>
-      <PrivateAssets>all</PrivateAssets>
-    </ProjectReference>
```
AArnott: That's not a duplicate. The first one is for **Package**Reference while the second is for **Project**Reference.

josetr: This may be a good time to check if the meassurements on my glasses have increased 😄.

PR #199, opened 2018-07-25
Implemented SetCloudBuildVariable for GitLab

ekwus:
Added the the capability for the versioning variables to be set as environment variables for use in the GitLab build runner environment.


ekwus:
Apologies, further testing shows that this isn't working the way its needed to work. The variables only exist within the process life time. I'll keep searching for a solution.

Issue #160, opened 2017-12-19
All packages in repo get version bump for every commit, even irrelevant ones


Tadimsky:
We have multiple projects inside a repo for all of our shared code that we use to build Nuget packages from (one Nuget package per project). I just started using this project for versioning all of our packages and really liking it so far - thanks for the awesome work!

The one thing I'd like to improve is how to get the Git Height working well for multiple projects.
Right now, if we make changes to one project (say 6 commits in one folder for the one project) then it will result in the version number being bumped up by 6 for all of our projects:
![image](https://user-images.githubusercontent.com/1149244/34175767-6c61f002-e4b2-11e7-8afa-570ead8c5af0.png)
This makes it hard to know when to update our project dependencies as it's unclear whether it's actually a new package or not.

Of course, this makes sense based on how the version is calculated, but I was wondering if there is a way to have this be based on the folder/project without having to move to separate git repos for each project. We currently have a `version.json` file per project so it would be neat if the git height could be calculated from the location of the `version.json` file down or something like that.
Any suggestions on how to do this better?

AArnott:
Good question. Yes, theoretically we could probably have a git height walk that is scoped to the path under the version.json file, so far I haven't done it, for this reason:

Files outside that path on your file system can still influence the packages. For example, if you have a Directory.Build.props file at the top level of your repo that spans your individual version.json files, and you change that top-level file, then all your individual packages _should_ be rev'd. Yet there's no way a scoped height counter would capture that.

You mention you don't know whether to update your project dependencies given the current system. Can you elaborate on that? Project references don't include version information, so what is there to update? 

I have several repos that build multiple packages and whenever we have a build we want to release, we just publish all the packages together. Yes, some of them may not actually have changes to them, but I've never had a customer complain yet.

Tadimsky:
Ah, yeah that does make sense with regards to changes above the project affecting the project...

Sorry, I mean that when consuming the updated Nuget packages in a different project (and repo), we now have updates for all of our packages even when only one package has actually been updated (based on code changes). It would be nice if there was someway of knowing when something actually changed.
Do you just bump to latest version for all packages then every time?

AArnott:
It varies. Sometimes I'll do what you say, by way of defining an MSBuild property with the version number in it and using that as an msbuild macro everywhere a `PackageReference` refers to it. That way I have just one place to update.

[...]

PR #223, opened 2018-09-13
[WIP] Allow skipping commits that don't contain changes relative to the project path.

danielcweber:
Consider the following scenario:

- I have a solution (.sln) with multiple projects (.csproj) in subfolders.
- These projects each produce a nuget package.
- A version.json is present in the solution-folder.
- On a build server (DevOps), always the complete solution is built, resulting in a new nupkg for each project, even if there were only changes in one of the projects.

I want to have a new package version only for the projects that have had changes. Having a separate version.json in each subfolder obviously won't work since the git height will still increase each time. This PR is a proof of concept and work in progress to enable skipping commits that don't contain any changes for the directory of the project, thus excluding them from contributing to the git height. This is enabled by setting `<SkipCommitsWithoutChangesInPath>true</SkipCommitsWithoutChangesInPath>` in MsBuild.

The PR is in early stages, there are changes in signatures of public methods and I'm sure I missed a lot of code paths so that the new feature won't work in all cases. Depending on whether this change is considered useful, I will continue work on this.

### DIFF
--- a/src/NerdBank.GitVersioning/GitExtensions.cs
+++ b/src/NerdBank.GitVersioning/GitExtensions.cs
@@ -45,7 +45,7 @@
                 baseVersion = VersionFile.GetVersion(commit, repoRelativeProjectDirectory)?.Version?.Version ?? Version0;
             }
 
-            int height = commit.GetHeight(c => CommitMatchesMajorMinorVersion(c, baseVersion, repoRelativeProjectDirectory));
+            int height = commit.GetHeight(null, c => CommitMatchesMajorMinorVersion(c, baseVersion, repoRelativeProjectDirectory));
             return height;
         }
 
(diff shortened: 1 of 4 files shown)


AArnott:
Thanks for contributing. Two quick thoughts, acknowledging I haven't yet seen your change:

1. Settings always go in the version.json file, since this will impact more than msbuild.
1. We probably want to consider cases where a change to a referenced project (transitively) also causes the version bump.

danielcweber:
>We probably want to consider cases where a change to a referenced project (transitively) also causes the version bump.

That's indeed not considered in my scenario. That adds some layers of complexity...

I'm considering whether this approach is so useful after all if cross-references are considered. Maybe it's advisable to split up solutions in that case.

AArnott:
> Maybe it's advisable to split up solutions in that case.

I appreciate your flexibility in thinking about how it could be implemented. That would indeed simplify it. However based on conversations I've had in #160, I think the primary use case includes projects within the same solution having unique version heights. Finding transitive P2Ps references for *all* projects is impossible, but for .NET SDK projects _might_ be possible, but as it would depend on MSBuild insights it may be undesirable anyway. So counting P2Ps may just be a requirement we can't meet. I'll ask for a vote on #160 to see how folks react.

japj:
So to summarize, we need a way to calculate which paths contribute to the git height.
1) for a fully isolated project (e.g. no project references), the project folder would be 'enough'
2) for a project with (ProjectReference) dependencies, the paths to the project references also contribute to the git height (and this might be recursive)
3) for projects that make use of "known MSBuild import hooks" (e.g. Directory.Build.props/targets) the folders with the import hooks contribute to the git height

2) and 3) will indeed depend on MSBuild insights.
Perhaps a "first" version of this mechanism could just contain a manual option to specify which (toplevel) folder contributes to git height. That way it would be an opt-in (for those "that know what they are doing").

A next version of this logic could then be prototyped/worked on to see how "correct" it can be made with actual MSBuild logic?

AArnott:
Ya, thinking about it more, I suspect a transitive walk along P2P refs is probably a non-starter for several reasons. Having the user explicitly specify the *set of folders* (allowing more than one) that contribute to the height is probably sufficient.

PR #237, opened 2018-10-22
Add semver2 support to npm

### LINKED ISSUE #236: npm setPackageVersion semver2 support
setPackageVersion is a nice helper function to update the version in a package.json file. Today it looks like it will always spit out a semver1 version. It would be nice if it could be parameterized to support semver2, as npm as semver2 support.

cdunford:
- An optional third paramater is added to setPackageVersion
  to specifiy the version of semver to use. The default value is 1
  to retain backwards compatibility.

fixes #236


cdunford:
I'm interested to know thoughts. I'm not terribly happy that in order to use semver2 you must provide all 3 parameters, even though others are optional. Perhaps a better solution would be to support a string or options object as the first parameter. If options are provided, they take precedence, even if a 2nd parameter is passed.

src/nerdbank-gitversioning.npm/ts/index.ts
```diff
@@ -61,13 +60,15 @@ export async function getVersion(projectDirectory?: string): Promise<IGitVersion
  * Sets an NPM package version based on the git height and version.json.
  * @param packageDirectory The directory of the package about to be published.
  * @param srcDirectory The directory of the source code behind the package, if different than the packageDirectory.
+ * @param semVer The version of semver to use in the package versioning.
  */
-export async function setPackageVersion(packageDirectory?: string, srcDirectory?: string) {
+export async function setPackageVersion(packageDirectory?: string, srcDirectory?: string, semVer: 1 | 2 = 1) {
```
AArnott: Rather than a parameter, can we follow the pattern we use for NuGet packages, by adding a field to version.json to indicate which semver version to use?

https://github.com/AArnott/Nerdbank.GitVersioning/blob/master/doc/versionJson.md

cdunford: Sounds good; I'll look into it. In the meantime, I've opened #238 which just includes the typescript upgrade, typings removal and some other minor cleanup changes.

AArnott:
This PR hasn't been updated in a month. Closing. I'll preparing the change I asked for myself. But I've added questions to the original issue that I need to get answers for.

PR #239, opened 2018-10-25
Fixes #198

### LINKED ISSUE #198: Reset computed version height when 3rd integer changes and height appears elsewhere
Hi,
we are using a versioning schema like 1.22.0-ci.{height} in the version.json.
We develop ci-versions in the dev branch. When we switch to a release branch we manually change the prerelease tag to rc (1.22.0-rc.{height}). We would expect the height to start at 0 again in the release branch but it continues from the value it had in dev.
Reading the docs it seems to be by design that the prerelease-tag is not taken into account for height-resets. Is it possible to achieve our intended behaviour? Shouldn't the prerelease-tag be taken into account at least when you are not using height for the patch-part but for the prerelease-part (like in our case)?

Thanks
Tobias

Kieranties:
Enables option to compare full version string when calculating height reset.  This resolves issue: https://github.com/AArnott/Nerdbank.GitVersioning/issues/198.

By adding the following property to `version.json`, the full version string is used for comparison during the calculation of the version height.  The option requires explicit enablement to prevent breaking changes.
```
{
  "compareFullVersion": true
}
```

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/VersionOracleTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/VersionOracleTests.cs
@@ -82,6 +82,32 @@ public class VersionOracleTests : RepoTestBase
         Assert.Equal(0, oracle.VersionHeightOffset);
     }
 
+    [Theory]
+    [InlineData("0.1", "0.2")]
+    [InlineData("0.1.0+{height}", "0.1.5+{height}")]
+    [InlineData("0.1.5-alpha0.{height}", "0.1.5-alpha1.{height}")]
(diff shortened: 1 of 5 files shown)


Kieranties:
@AArnott Do you have any feedback/guidance on this PR?  I'm looking to start implementing builds soon and would love to have an official build supporting this.

AArnott:
There are some changes left to be made so that the answer is the same no matter which API is used. I'm working through it.

AArnott:
I found a few inconsistencies around our support for `{height}` to appear anywhere other than the build number (3rd integer) position. I'm working my way through it, with tests.

Kieranties:
Great stuff!

PR #253, opened 2018-11-13
Enable '-' to be used to split height in prerelease label

Kieranties:
A small adjustment to the prerelease regex to enable the format `1.0.0.0-alpha1-{height}` to be used as well as the existing format of `1.0.0-alpha1.{height}`


Kieranties:
I've just performed some additional testing and can see that the format '1.0.0.0-alpha1-{height}' always causes and error when running.  However if using `1.0.0-alpha1.{height}`, using the Semver1 property returns a correctly formatted value.

I think this PR can be ignored.

Issue #149, opened 2017-09-29
[Enhance] FSharp support


bradphelan:
When added to an fsharp project there is an error about CodeDom missing. Not surprising but is there a chance support can be added?

bradphelan:
Actually the error is weird. I have some fsharp projects but Nerdbank.Gitversioning is not installed. I get

    676 Warning(s)
    0 Error(s)

> Time Elapsed 00:03:11.37
> .\build.bat : Building WeinCad.Net.sln failed with exitcode 1.
> At line:1 char:1
> + .\build.bat 2>&1 | Tee-Object -FilePath build.log
> + ~~~~~~~~~~~~~~~~
>     + CategoryInfo          : NotSpecified: (Building WeinCa...ith exitcode 1.:String) [], RemoteException
>     + FullyQualifiedErrorId : NativeCommandError
> 
> : C:\Users\phelan\workspace\WeinCAD.NET\packages\Nerdbank.GitVersioning.2.0.41\build\Nerdbank.GitVersioning.targets(135,5): CodeDomProvider not available for language: F#. No version info will be embedded i
> nto assembly.
> Stopped due to error
> C:\Users\phelan\workspace\WeinCAD.NET [feature/general-z-axis-corrections ↑ +56 ~84 -0 !]> gvim .\build.log
> 

but it doesn't say which assembly and there is nothing else in the log

AArnott:
NB.GV already has some support for manually generating source code when CodeDom is absent, since this is required for `dotnet build`. So yes, I think F# support could be added.
As I'm not an F# syntax expert, I'd prefer to add this by way of accepting someone else's PR.

bradphelan:
Ahhhh. Ignore my previous post about the error still being there even though the FSharp projects didn't include the nuget package. That was some weird stuff that hasn't happened again. But I created a pull request anyway to at least put the name of the assembly in the error messages.  :)

I might add stuff for FSharp if it turns out I need them.

PR #255, opened 2018-11-15
Fix for issue #149 - Adding F# support for AssemblyInfo generation

kwpatrick:
Adding F# support involved some minor changes to how the assembly infos are generated, since the NET461 target now needs to be able to try both CodeDom and native code generation paths.  This resolves issue #149.

One thing to note: F# requires code to be in a module or namespace, and so the new F# path will apply a namespace (the provided root namespace + "Properties" if possible, "AssemblyInfo.Properties" if nothing is provided.  Other code generation targets should be unaffected by this.


AArnott:
Thanks, @kwpatrick. I made a few mainly cosmetic changes, but I like what you've done here.

PR #257, opened 2018-11-25
Use SemVer 2.0 for NPM packages

### LINKED ISSUE #236: npm setPackageVersion semver2 support

AArnott:
This has been around and supported for at least six years in NPM.

Fixes #236


cdunford:
Sorry. This was still on my radar to do, but great to see you've done it. Thanks!

AArnott:
No problem.

PR #276, opened 2018-12-14
Add NuGet-semver v1 compatible version as ChocolateyPackageVersion

### LINKED ISSUE #275: Cannot upgrade to 2.3 due to breaking changes (removal of  -g)
>So if your scenario requires both a NuGet-compatible semver v1 value for Chocolatey and a semver 2.0 version stamp on your actual NuGet package, NB.GV 2.3 is going to create problems for you unless we somehow fill that gap.

This is exactly my scenario as Chocolatey requires NuGet SemVer 1 compatible versions and gets tripped up by tags that start with a number. In my project, I build a nuget package and a chocolatey package and I'd like to use SemVer 2 for NuGet while having a "nuget semver 1" string available for chocolatey.

AArnott:
- Add ChocolateyPackageVersion to VersionOracle
- Define ChocolateyPackageVersion as an MSBuild property

Fixes #275

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -945,16 +945,25 @@ public class BuildIntegrationTests : RepoTestBase
         Assert.Equal(versionOptions.Version.Prerelease, buildResult.PrereleaseVersion);
         Assert.Equal(expectedBuildMetadata, buildResult.SemVerBuildSuffix);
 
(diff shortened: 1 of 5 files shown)


AArnott:
@onovotny Will this suit you?

onovotny:
Does that make it into the cloud vars, so we can use `NBGV_ChocolateyPackageVersion`?

AArnott:
Yes.

onovotny:
Should be perfect then, thanks!

PR #278, opened 2018-12-15
Pipelines now supports artifacts and versions from forks

onovotny:
This will enable greater testing from PR's


onovotny:
It was in Sprint 142? Not sure the specific release notes.

AArnott:
Can you send a PR to update [the docs](https://github.com/AArnott/Nerdbank.GitVersioning/blob/master/doc/cloudbuild.md#set-the-cloud-build-number-as-a-build-step) to point to a sample that no longer has the isfork check as well?

onovotny:
Will do shortly

PR #291, opened 2019-01-27
Add "prepare-release" command

### LINKED ISSUE #287: Feature suggestion: nbgv command to create a release branch
For some of my projects, I use the following branching model in conjunction with `Nerdbank.GitVersioning`:

- `master` always contains the latest development version, releases are built from a `release/vX.Y` branch
- `master` is always a prerelease version, so in `version.json` the version is set to e.g. `1.2-pre`
- When a release branch is created, it is branched off from master and the prerelease tag is removed, while the version of master is increments. So in the above example, `master` would switch to `1.3-pre` and a release branch `release/v1.2`would be created where the version is just `1.2`

I wrote a small script to automate this process. However, I think this would be useful to be included as a command in the `nbgv` tool.

This way, creating a new release would be as simple as running

```
nbgv create-release
```
which would

- read the version from `version.json`
- create a release branch
- edit and commit new `version.json` on release branch
- edit and commit new `version.json` on master branch

Would you be open to include such a feature?  
I would be happy to implement it and provide a PR.

ap0llo:
This PR adds the `prepare-release` command as discussed in #287 

I made a minor modification to the feature compared to what was discussed in the issue. The setting to control the name of release branches does not use `{0}` as a placeholder for the version but `{version}`. I think this makes it more consistent with the version setting which uses `{height}`

Closes #287

### DIFF
--- a/src/NerdBank.GitVersioning/GitExtensions.cs
+++ b/src/NerdBank.GitVersioning/GitExtensions.cs
@@ -387,7 +387,7 @@
             Requires.NotNullOrEmpty(pathUnderGitRepo, nameof(pathUnderGitRepo));
             var gitDir = FindGitDir(pathUnderGitRepo);
 
-            if(useDefaultConfigSearchPaths)
+            if (useDefaultConfigSearchPaths)
             {
                 // pass null to reset to defaults
                 GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, null);
@@ -399,7 +399,7 @@
                 GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, string.Empty);
                 GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.System, string.Empty);
             }
-            
+
             return gitDir == null ? null : new Repository(gitDir);
         }
 
--- a/src/NerdBank.GitVersioning/GitExtensions.cs
+++ b/src/NerdBank.GitVersioning/GitExtensions.cs
@@ -380,16 +380,26 @@
         /// Opens a <see cref="Repository"/> found at or above a specified path.
         /// </summary>
         /// <param name="pathUnderGitRepo">The path at or beneath the git repo root.</param>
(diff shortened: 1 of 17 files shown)


src/NerdBank.GitVersioning/ReleaseManager.cs
```diff
@@ -0,0 +1,255 @@
...
+                throw new ReleasePreparationException(ReleasePreparationError.VersionDecrement);
+            }
+
+            this.stdout.WriteLine($"Setting version to {newVersion}");
+
+            versionOptions.Version = newVersion;
+            var filePath = VersionFile.SetVersion(projectDirectory, versionOptions);
+            
+            Commands.Stage(repository, filePath);
+            repository.Commit($"Set version to '{versionOptions.Version}'", signature, signature, new CommitOptions() { AllowEmptyCommit = true } );
```
AArnott: Is there a reason we want to create an empty commit rather than just skipping the step?

AArnott: I'm assuming not, so I've changed this.

AArnott:
I've fixed the `$schema` issue.

src/NerdBank.GitVersioning/GitExtensions.cs
```diff
@@ -380,16 +380,26 @@ public static string FindLibGit2NativeBinaries(string basePath)
         /// Opens a <see cref="Repository"/> found at or above a specified path.
         /// </summary>
         /// <param name="pathUnderGitRepo">The path at or beneath the git repo root.</param>
+        /// <param name="useDefaultConfigSearchPaths">Specifies whether to use default settings for looking up global and system settings.</param>
```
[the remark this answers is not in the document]
AArnott: Thanks! I imagine I originally cleared config search paths for unit test stability. I recall I had a problem at one time where the tests would pass for me and fail on an appveyor machine because it had no user-level git configuration so commits would fail. I believe your PR is the first one where nbgv actually authors a commit on behalf of the user, which certainly makes sense to bring in machine configuration. So thanks for fixing that. :)

PR #296, opened 2019-02-02
fix and re-enable NativeVersionInfo_CreateNativeResourceDll for AzurePipelines

japj:
Fixes #294

PR #297, opened 2019-02-02
Reapply "Merge pull request #296 from japj/fix294"

AArnott:
This reverts commit e442089d828ef60c2cc1e5ce2283214b0614c107.

Reapplies #296


japj:
Ok, apparently I cannot reproduce it from visualstudio, but I did reproduce it (only once) from cli with
`dotnet test src/Nerdbank.GitVersioning.Tests  -c Release --filter "TestCategory!=FailsOnAzurePipelines"`

So this might actually be a race condition somewhere.

PR #298, opened 2019-02-03
persist msbuild files generated during test before actual execution

japj:
The testProject and imported targets files were not persisted during testing.

I believe this might be the cause of the unstable tests, as the build engine actually will sometimes error out on missing files during testing like:
`Microsoft.Build.Exceptions.InvalidProjectFileException : The imported project "C:\Users\VssAdministrator\AppData\Local\Temp\b0hpof3h.oba\NerdBank.GitVersioning.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.  C:\Users\VssAdministrator\AppData\Local\Temp\b0hpof3h.oba\projdir\test.prj`


AArnott:
It worked. I've completed #297 which contains this change.

japj:
yeah, I think there must be some weirdness with MSBuild integration going on for it to decide to not use the in-memory-only project files, but that is way too much out of my knowledge area at the moment.

Issue #300, opened 2019-02-04
Prepare-release command support for incrementing build variable


josesimoes:
I've used this new command a dozen times today to publish some releases and it's working great.
Great improvement that really speeds up this process!! 💯 

A questions: why shouldn't the versionIncrement field allow to use "build" and it's limited to major & minor? Sometimes one really wants to increment _just_ the build field.

AArnott:
Thanks for the feedback. @ap0llo any particular reason you didn't implement support for incrementing the build variable?

ap0llo:
I mainly omitted it because I didn't really see a use case, but I don't see why support for it couldn't be added.
The only aspect that could tricky is the naming of the release branch as this currently only cosiders major and minor version.

@josesimoes In your use case, would you expected to have release branches for two versions that only differ in the build number?

ap0llo:
I'll look into it on the weekend

josesimoes:
That's great!

Answering your question above, yes, sometimes we do have intermediate releases with quick fixes and such that only bump the build number.

josesimoes:
Thinking about the release workflow it would be very helpfull if there was a switch in this command that allowed you to specify what to bump: major or minor.
The default would be what's in the config file but it does help to be able to specify that from the command line without having to manually change the file (which would kind of defeat the purpose of the command, right?)

Think of this: next time you are releasing a major version. You would have to edit version.jomson for either changing the version or the bump setting. 

Am I making sense?

ap0llo:
Yes, makes sense. What you can do today is to override the automatically calculated next version wihtout changing version.json:

```
nbgv prepare-release --nextVersion X.Y
```

josesimoes:
@ap0llo got you!

I just tried that and that is kind of messing with the version.json, but I guess that's expected.

starting point:
```
...
"1.1.0-preview.{height}",
...
```

command: `nbgv prepare-release --nextVersion 1.3` 

result:
```
...
"version": "1.3",
...
```

I have to use: `nbgv prepare-release --nextVersion 1.3.0-preview.{height}` in order to achieve the "expected" outcome (meaning: following the version pattern that was being used).

AArnott:
I would have expected `-alpha` (or whatever the 'default' prerelease tag set in version.json is) to be added automatically, along with any subsequent prerelease tags in the original.

ap0llo:
[...]

PR #301, opened 2019-02-09
Allow incrementing the "build" of the version in the prepare-release command

ap0llo:
As discussed in #300, this PR add support for incrementing the "build" of a version when running `prepare-release`.

This can be achieved by setting the `versionIncrement` setting in `version.json` to `build`. 
The command will abort with an error if the current version only has a major and minor version.

### DIFF
--- a/src/NerdBank.GitVersioning/ReleaseManager.cs
+++ b/src/NerdBank.GitVersioning/ReleaseManager.cs
@@ -58,7 +58,7 @@
             DetachedHead,
 
             /// <summary>
-            /// The versionIncrement setting cannot be applied to the current version
+            /// The versionIncrement setting cannot be applied to the current version.
             /// </summary>
             InvalidVersionIncrementSetting,
         }
--- a/src/NerdBank.GitVersioning/ReleaseManager.cs
+++ b/src/NerdBank.GitVersioning/ReleaseManager.cs
@@ -56,6 +56,11 @@
             /// HEAD is detached. A branch must be checked out first.
             /// </summary>
             DetachedHead,
+
+            /// <summary>
+            /// The versionIncrement setting cannot be applied to the current version
+            /// </summary>
+            InvalidVersionIncrementSetting,
         }
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


ap0llo:
Interesting, there seems to be some change on the Azure DevOps side, as this error also happens when building the latest commit on master (b668f0d2bea3e56c87e6f3b48b1b082c9735638f) 

Building the NPM package fails with 
```
Error:
An assembly specified in the application dependencies manifest (nbgv.deps.json) was not found:
     package: 'System.Security.Cryptography.ProtectedData', version: '4.3.0'
     path: 'runtimes/win/lib/netstandard1.3/System.Security.Cryptography.ProtectedData.dll'
```

src/NerdBank.GitVersioning/VersionOptions.cs
```diff
@@ -1139,7 +1139,12 @@ public enum ReleaseVersionIncrement
             /// <summary>
             /// Increment the minor version after creating a release branch
             /// </summary>
-            Minor
+            Minor,
+
+            /// <summary>
+            /// Inrementn the build number (the third number in a version) after creating a release branch
```
AArnott: typo: Inrementn 
```suggestion
            /// Increment the build number (the third number in a version) after creating a release branch.
```

src/NerdBank.GitVersioning/ReleaseManager.cs
```diff
@@ -56,6 +56,11 @@ public enum ReleasePreparationError
             /// HEAD is detached. A branch must be checked out first.
             /// </summary>
             DetachedHead,
+
+            /// <summary>
+            /// The versionIncrement setting cannot be applied to the current version
```
AArnott: nit: trailing period.

```suggestion
            /// The versionIncrement setting cannot be applied to the current version.
```

src/NerdBank.GitVersioning/version.schema.json
```diff
@@ -142,13 +142,13 @@
             "branchName": {
               "description": "Defines the format of release branch names. Format must include a placeholder '{version}' for the version",
               "type": "string",
-              "pattern": ".*\\{0\\}.*",
+              "pattern": ".*\\{version\\}.*",
```
AArnott: Is this a breaking change from the feature when first introduced, or just updating the schema to match what the code already expected?

AArnott:
> Interesting, there seems to be some change on the Azure DevOps side

I see what you mean. I'm looking into it. Azure Pipelines did indeed update the VS version on their agents and my attempt to pin the .NET SDK I've learned recently had some holes in it. So I'll see if I can fix that up and restart the build on your PR.

src/NerdBank.GitVersioning/version.schema.json
ap0llo: The latter, matching what the code already expected. During the original implementation I changed the placeholder from `{0}` to `{version}` and I forgot to adjust the schema accordingly.

AArnott:
I made a fix, and it got master building, and my own merge of your PR passes the build. I don't know why the build fails when coming from the github PR. I'll merge manually if necessary after you make the couple fixes I asked for.

ap0llo:
I applied your changes, and your fix seems to work, the build passed 😃

PR #307, opened 2019-03-11
Add GeneratedCodeAttribute to ThisAssembly

JohannesWill:
Add GeneratedCodeAttribute to ThisAssembly

### DIFF
--- a/src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs
+++ b/src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs
@@ -413,7 +413,9 @@
 
             internal override void StartThisAssemblyClass()
             {
-                this.codeBuilder.AppendLine("do()\r\ntype internal ThisAssembly() =");
+                this.codeBuilder.AppendLine("do()");
+                this.codeBuilder.AppendLine($"[<System.CodeDom.Compiler.GeneratedCode(\"{ThisAssembly.AssemblyName}\",\"{ThisAssembly.AssemblyVersion}\")>]");
+                this.codeBuilder.AppendLine("type internal ThisAssembly() =");
             }
         }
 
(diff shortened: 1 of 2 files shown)


AArnott:
Thanks for contributing.

Why is this a worthwhile attribute to include? I'm not too familiar with this attribute.

JohannesWill:
The GeneratedCodeAttribute class can be used by code analysis tools to identify computer-generated code, and to provide an analysis based on the tool and the version of the tool that generated the code.

see 
https://docs.microsoft.com/en-us/dotnet/api/system.codedom.compiler.generatedcodeattribute?view=netframework-4.7.2

AArnott:
Interesting. We have several code generators. You've only updated the C# one. We have other languages, and [codedom-based one](https://github.com/AArnott/Nerdbank.GitVersioning/blob/166a3e3bd611732b5f451266a27d9c8185bffbe7/src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs#L118). Can you update them all so they're consistent?

src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs
```diff
@@ -119,8 +119,14 @@ private CodeTypeDeclaration CreateThisAssemblyClass()
             {
                 IsClass = true,
                 IsPartial = true,
-                TypeAttributes = TypeAttributes.NotPublic | TypeAttributes.Sealed,
+                TypeAttributes = TypeAttributes.NotPublic | TypeAttributes.Sealed | TypeAttributes.Abstract,
```
AArnott: Curious: Why was `Abstract` a required addition?
Does it make sense to be both abstract *and* sealed? (perhaps that's the way C# "static" classes are actually implemented?)

JohannesWill: It was for the C# "static" classes 
I removed it.

AArnott: Thanks.  I didn't know you could create static classes with CodeDOM. Nice trick. :)

AArnott:
Thanks for contributing!

PR #310, opened 2019-03-20
Add Travis CI support

qmfrederik:
This adds cloud provider support for Travis CI.

### DIFF
--- a/src/NerdBank.GitVersioning/CloudBuild.cs
+++ b/src/NerdBank.GitVersioning/CloudBuild.cs
@@ -19,7 +19,7 @@
             new AtlassianBamboo(), 
             new Jenkins(),
             new GitLab(),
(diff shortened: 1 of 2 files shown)


src/NerdBank.GitVersioning/CloudBuild.cs
```diff
@@ -19,6 +19,7 @@ public static class CloudBuild
             new AtlassianBamboo(), 
             new Jenkins(),
             new GitLab(),
+            new Travis()
```
AArnott: Nit: add a trailing comma so the next addition, like yours, doesn't edit existing lines.

qmfrederik: Done, thanks for the feedback!

PR #279, opened 2018-12-15
Add a .NET Core 3 target for the global tool

onovotny:
Fixes #277

PR #315, opened 2019-04-17
add netcoreapp3.0 global tool

ctaggart:
Based on #279 from @onovotny, but uses latest preview 3 and sets the global.json just on the Azure Pipelines build.


azure-pipelines.yml
```diff
@@ -35,12 +35,16 @@ jobs:
     displayName: Set build number
 
   - task: DotNetCoreInstaller@0
-    displayName: Install .NET Core SDK 2.1.500
+    displayName: Install .NET Core SDK
     inputs:
       packageType: sdk
-      version: 2.1.500
+      version: 3.0.100-preview3
+  - script: dotnet new globaljson --sdk-version 3.0.100-preview3
```
onovotny: Neat. you can also use `3.0.100-preview` as the version as the lookup rules will find the highest matching preview. That means you won't forget to update it later.

ctaggart:
I'm not sure what this error is from:

image

I know that switching from TargetFramework to TargetFrameworks adds an additional folder. May be something in the gulpfile needs updating.

src/nbgv/nbgv.csproj
```diff
@@ -4,7 +4,7 @@
     <ToolCommandName>nbgv</ToolCommandName>
     <PackAsTool>True</PackAsTool>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp3.0</TargetFrameworks>
```
ctaggart: If you are missing one of the target frameworks, does it still load? Hoping so.

src/Directory.Build.props
```diff
@@ -1,8 +1,6 @@
 <Project>
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)..\obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
-    <OutputPath>$(MSBuildThisFileDirectory)..\bin\$(MSBuildProjectName)\$(Configuration)\</OutputPath>
```
ctaggart: This combined with TargetFrameworks does not work. It is easier to just `dotnet publish`.

src/nbgv/nbgv.csproj
```diff
@@ -4,14 +4,12 @@
...
     <RootNamespace>Nerdbank.GitVersioning.Tool</RootNamespace>
     <Description>A .NET Core Tool that can install, read and set version information based on git history, using Nerdbank.GitVersioning.</Description>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="NuGet.PackageManagement" Version="4.9.3" NoWarn="NU1701" />
-    <PackageReference Include="NuGet.Resolver" Version="4.9.3" />
```
ctaggart: These were only used by `GetLatestPackageVersionAsync`. They are netframework only and cause a bunch of dependency problems. Hopefully there is an easier way with NuGet 5.0.

azure-pipelines.yml
ctaggart: This doesn't work because of the `netcoreapp3.0` dependency, so I removed it.

Issue #314, opened 2019-04-17
LibGit2Sharp not working on sdk:2.2.105-alpine3.9


ctaggart:
Steps to reproduce:
```
docker run --rm -it mcr.microsoft.com/dotnet/core/sdk:2.2.105-alpine3.9 sh
dotnet tool install -g nbgv
$HOME/.dotnet/tools/nbgv get-version
```

Error:
```
/ # $HOME/.dotnet/tools/nbgv get-version

Unhandled Exception: System.TypeInitializationException: The type initializer for 'LibGit2Sharp.Core.NativeMethods' threw an exception. ---> System.DllNotFoundException: Unable to load shared library 'git2-a904fc6' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: Error loading shared library libgit2-a904fc6: No such file or directory
   at LibGit2Sharp.Core.NativeMethods.git_libgit2_init()
   at LibGit2Sharp.Core.NativeMethods.InitializeNativeLibrary()
   at LibGit2Sharp.Core.NativeMethods..cctor()
   --- End of inner exception stack trace ---
   at LibGit2Sharp.Core.NativeMethods.git_libgit2_opts(Int32 option, UInt32 level, String path)
   at LibGit2Sharp.GlobalSettings.SetConfigSearchPaths(ConfigurationLevel level, String[] paths)
   at Nerdbank.GitVersioning.GitExtensions.OpenGitRepo(String pathUnderGitRepo, Boolean useDefaultConfigSearchPaths)
   at Nerdbank.GitVersioning.Tool.Program.OnGetVersionCommand(String projectPath, String format, String singleVariable, String versionOrRef) in D:\a\1\s\src\nbgv\Program.cs:line 261
   at Nerdbank.GitVersioning.Tool.Program.Main(String[] args) in D:\a\1\s\src\nbgv\Program.cs:line 131
Aborted
```

ctaggart:
Works fine with `sdk:2.2.105-alpine3.8`.

ctaggart:
Hoping this is fixed by #316

ctaggart:
This does not appear to have been fixed by #316 😢 . Here is how I tested locally:
```
docker run -it --rm -v ${PWD}:/src -w /src mcr.microsoft.com/dotnet/core/sdk:2.2.105-alpine3.9 sh
[...]

PR #316, opened 2019-04-18
add latest runtime identifiers

ctaggart:
Same map as https://github.com/dotnet/sourcelink/pull/263

I noticed this drops the `corert` identifiers. Were those in the original import?


AArnott:
Yes, those `corert` identifiers were present originally in the place I copied this from, but evidently [they removed them last December](https://github.com/dotnet/sourcelink/commit/d5ce73f7e782beb135376ea50d08d8448b21c263#diff-74c40ba3b1b98b0a00c3b4f9024180b9). Personally I have no reason to believe they are important for this library.

I see from #314 you're hoping this will start working on Alpine. Any reason not to add an Alpine 3.9 test in the azure-pipelines.yml file as part of this change by following this template?
https://github.com/AArnott/Nerdbank.GitVersioning/blob/01ebb81f18af1ce64659ad06267d86be96d1f8ad/azure-pipelines.yml#L133-L139

AArnott:
Thanks, @ctaggart. Given the failure, I tried `docker run` of that same image locally and it just exited immediately. `/bin/bash` isn't present. Do you have any idea how to use/test this image so we can get this working?

AArnott:
That's fine. We can just remove it. But would you mind removing the alpine test commit and your reverting commit from the PR before I complete it?

ctaggart:
I'll try to test it locally now. Does the squash merge button work well for keeping a clean history on `master`?

AArnott:
Sure. I'm happy to squash. I don't usually do it in case you're building on your branch in your fork for other things.

PR #323, opened 2019-04-22
Emit a warning when SemVer 2.0 format version is used with explicit SemVer 1.0

### LINKED ISSUE #318: Emit a warning when SemVer 2.0 format version is specified with explicitly selecting SemVer 1.0.
The following seems potentially worthy of a warning:

```
"version": "1.0.0-preview.{height}",
  "nugetPackageVersion": {
    "semVer": 1
  },
 ```

robmen:
Should the user explicitly set `nugetPackageVersion.semVer` to `1` but
use a SemVer 2 prerelease in their `version` (such as: `1.0.0-preview.{height})
a warning will now be displayed telling the user they are probably not
getting what they expected.

Fixes #318


robmen:
We opened the issue, figured we should help try to fix it. :smile:

PR #324, opened 2019-04-22
nbgv for dotnet 3.0

### LINKED ISSUE #277: Add netcoreapp3.0 target for nbgv
.NET Core global tools will auto roll-forward to the next minor. So for people that only have `netcoreapp2.2`, the current `netcoreapp2.1` target will work just fine. Major versions do not roll forward, so if a user only has `netcoreapp3.0`, then the tool will fail. 

The solution is to either have users with both a .NET Core 2.x SDK and a 3.x SDK or to have tools multi-target. For ease of users, I'm recommending all .NET Core global tools add a `netcoreapp3.0` target.

/cc @kathleendollard

ctaggart:
Squash merged #315. Closes #277.


build.ps1
```diff
@@ -15,25 +15,36 @@ Param(
     [string]$MsBuildVerbosity = 'minimal'
 )
 
-$msbuildCommandLine = "msbuild `"$PSScriptRoot\src\Nerdbank.GitVersioning.sln`" /m /verbosity:$MsBuildVerbosity /nologo /p:Platform=`"Any CPU`" /t:build,pack"
+$msbuildCommandLine = "dotnet build `"$PSScriptRoot\src\Nerdbank.GitVersioning.sln`" /m /verbosity:$MsBuildVerbosity /nologo /p:Platform=`"Any CPU`""
+$msbuildPack = "dotnet pack --no-build `"$PSScriptRoot\src\Nerdbank.GitVersioning.sln`" -o bin"
```
AArnott: Why do we need this `-o bin` parameter? Can't we change the msbuild authoring if necessary to get it to naturally build to the right folders?

src/Directory.Build.props
```diff
@@ -1,8 +1,6 @@
 <Project>
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)..\obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
-    <OutputPath>$(MSBuildThisFileDirectory)..\bin\$(MSBuildProjectName)\$(Configuration)\</OutputPath>
```
AArnott: Uh what? Doesn't this eliminate my readonly source tree and put bin/obj folders back under source project directories? I really don't want that. Why is that necessary as part of building another .netcore target for nbgv?

src/global.json
```diff
@@ -1,5 +0,0 @@
-{
-  "sdk": {
-    "version": "2.1.500"
```
AArnott: I'm not a fan of this change either. I like to 'pin' my SDK versions so that over time, as Microsot (inevitably) breaks compatiblity with build tools, I can always come back and build something even years later.
Why was this necessary?

src/nbgv/Program.cs
```diff
@@ -639,34 +633,9 @@ private static ExitCodes OnPrepareReleaseCommand(string projectPath, string prer
...
-            var pkg = await NuGetPackageManager.GetLatestVersionAsync(
-                packageId,
-                framework,
-                resolutionContext,
-                sourceRepository,
-                NullLogger.Instance,
-                cancellationToken);
-
-            return pkg.LatestVersion.ToNormalizedString();
+           return Task.FromResult("2.3.138"); // TODO
```
AArnott: I guess you're coming back to this?

src/nbgv/nbgv.csproj
```diff
@@ -4,14 +4,12 @@
     <ToolCommandName>nbgv</ToolCommandName>
     <PackAsTool>True</PackAsTool>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp3.0</TargetFrameworks>
```
AArnott: Now this part makes sense and looks relevant. :)

src/global.json
ctaggart: I like to pin the SDKs too. I just `git mv`ed it up to the root folder. When I run `build.ps1` locally or run command from the root folder, I want them using the same dotnet.

src/nbgv/Program.cs
ctaggart: `NuGet.PackageManagement` and `NuGet.Resolver` and .NET Framework only. You have it pinned to 4.9.3, which is net46. https://www.nuget.org/packages/NuGet.PackageManagement/4.9.3
Are you sure you want these dependencies? I can attempt putting the dependencies and code back to verify if it works or not with them.

build.ps1
ctaggart: `-o bin` is where to put the *.nupkg packages. The [CopyFiles task](https://github.com/ctaggart/Nerdbank.GitVersioning/blob/5513a59a28417e729f5c92a10183ae17de2ea2ca/azure-pipelines.yml#L106-L108) looks for them there.

ctaggart: That is where they were going before as well.

src/Directory.Build.props
ctaggart: I'm not I understand the desire not to just use the defaults here. The important part of this PR is that the npm wrapper needs to use the output from a `dotnet publish` for the `deps.json` to be correct. Somewhere in the above settings, `$(TargetFramework)` may be needed.

global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "3.0.100-preview4-011223"
```
ctaggart: Still pinned, just moved to root.

src/global.json
AArnott: Ah, ok good. I only discovered a short while ago that dotnet.exe uses the *current directory* to search for global.json rather than the directory of the thing being built. So in other places I have changed the scripts to set the working directory to the `src` folder before invoking `dotnet`, but took a mental note to move `global.json` up to the top-level folder in the future. So you're just taking care of it for me. :)

azure-pipelines.yml
```diff
@@ -59,7 +60,10 @@ jobs:
     inputs:
       ProjectDirectory: 'src/nerdbank-gitversioning.npm'
 
-  - script: dotnet build -c $(BuildConfiguration) --no-restore /t:build,pack /bl:"$(Build.ArtifactStagingDirectory)/build_logs/msbuild.binlog"
+  - script: |
+      dotnet build -c $(BuildConfiguration) --no-restore /bl:"$(Build.ArtifactStagingDirectory)/build_logs/msbuild.binlog"
+      dotnet pack -c $(BuildConfiguration) --no-build -o ..\bin
+      dotnet publish -c $(BuildConfiguration) --no-build .\nbgv\nbgv.csproj -f netcoreapp2.1 -o .\nerdbank-gitversioning.npm\out\nbgv.cli\tools\netcoreapp2.1\any
```
ctaggart: This is needed in order to create a `nbgv.deps.json` that works. A different folder can be published to.

AArnott:
Why do you close these PRs only to open new ones repeatedly? This makes the 3rd PR I think.

PR #329, opened 2019-04-26
Add a GitCommitDate property to ThisAssembly

qmfrederik:
Adds a `ThisAssembly.GitCommitDate` property, which contains the Git commit date from HEAD.

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -743,6 +743,18 @@ public class BuildIntegrationTests : RepoTestBase
         Assert.Equal(result.AssemblyCopyright, thisAssemblyClass.GetField("AssemblyCopyright", fieldFlags)?.GetValue(null));
         Assert.Equal(result.GitCommitId, thisAssemblyClass.GetField("GitCommitId", fieldFlags)?.GetValue(null) ?? string.Empty);
 
+        if (gitRepo)
+        {
+            Assert.True(long.TryParse(result.GitCommitDateTicks, out _), $"Invalid value for GitCommitDateTicks: '{result.GitCommitDateTicks}'");
+            DateTimeOffset gitCommitDate = new DateTimeOffset(long.Parse(result.GitCommitDateTicks), TimeSpan.Zero);
+            Assert.Equal(gitCommitDate, thisAssemblyClass.GetProperty("GitCommitDate", fieldFlags)?.GetValue(null) ?? string.Empty);
+        }
+        else
+        {
+            Assert.Empty(result.GitCommitDateTicks);
+            Assert.Null(thisAssemblyClass.GetProperty("GitCommitDate", fieldFlags));
+        }
+
         // Verify that it doesn't have key fields
         Assert.Null(thisAssemblyClass.GetField("PublicKey", fieldFlags));
         Assert.Null(thisAssemblyClass.GetField("PublicKeyToken", fieldFlags));
@@ -939,6 +951,7 @@ public class BuildIntegrationTests : RepoTestBase
         Assert.Equal(idAsVersion.Build.ToString(), buildResult.BuildVersionNumberComponent);
         Assert.Equal($"{idAsVersion.Major}.{idAsVersion.Minor}.{idAsVersion.Build}", buildResult.BuildVersionSimple);
         Assert.Equal(this.Repo.Head.Commits.First().Id.Sha, buildResult.GitCommitId);
+        Assert.Equal(this.Repo.Head.Commits.First().Author.When.UtcTicks.ToString(), buildResult.GitCommitDateTicks);
         Assert.Equal(commitIdShort, buildResult.GitCommitIdShort);
         Assert.Equal(versionHeight.ToString(), buildResult.GitVersionHeight);
         Assert.Equal($"{version.Major}.{version.Minor}", buildResult.MajorMinorVersion);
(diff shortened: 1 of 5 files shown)


src/NerdBank.GitVersioning/VersionOracle.cs
```diff
@@ -256,6 +257,11 @@ public IEnumerable<string> BuildMetadataWithCommitId
         /// </summary>
         public string GitCommitIdShort { get; }
 
+        /// <summary>
+        /// Gets the Git revision control commit date for HEAD (the current source code version).
+        /// </summary>
+        public string GitCommitDate { get; }
```
AArnott: I think we should define this as a `DateTime`.

src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs
```diff
@@ -310,6 +313,7 @@ private void GenerateThisAssemblyClass()
                     { "AssemblyCompany", this.AssemblyCompany },
                     { "AssemblyConfiguration", this.AssemblyConfiguration },
                     { "GitCommitId", this.GitCommitId },
+                    { "GitCommitDate", this.GitCommitDate },
```
AArnott: Can we make this a `DateTime`, initialized using one of the integer based ctor's like [this one](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.-ctor?view=netcore-2.2#System_DateTime__ctor_System_Int64_System_DateTimeKind_)?
It seems like a `DateTime` property on the `ThisAssembly` class would be more generally useful (and faster to process, more flexibly formatted) than a datetime that is pre-formatted in a particular way.

We'll have to "tunnel" the ctor args as a string for MSBuild interop, but that could just be an int64 formatted as a base10 string, for example.

qmfrederik:
@AArnott So, I updated the PR to make sure the date is typed as a DateTimeOffset in the VersionOracle class, and a string representation of a long (the number of ticks since 1/1/0001) in the MSBuild tasks.

The AssemblyInfo file itself would need to invoke the constructor of DateTimeOffset with a const, something the code generators can't express yet.

I'll try to work on that in the coming days.

qmfrederik:
@AArnott OK - I converted the `GitCommitDate` to a read-only property of type `DateTimeOffset`. Can you have a look at the diff and let me know if you like the direction this is going?

Couple of questions:
- It looks like the .NET code generators are used when building on .NET Framework and custom code generators when building on .NET Core. I'm not sure the .NET Framework code is being tested in CI. Is there a reason to keep both branches, or would it be OK to enable the .NET Core code path on .NET Framework as well?
- I'm not sure the F# code generator is being tested. I don't know much about F#, is there a way to add F# equivalent of BuildIntegrationTests.AssemblyInfo?

src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
```diff
@@ -937,6 +949,7 @@ private void AssertStandardProperties(VersionOptions versionOptions, BuildResult
         Assert.Equal(idAsVersion.Build.ToString(), buildResult.BuildVersionNumberComponent);
         Assert.Equal($"{idAsVersion.Major}.{idAsVersion.Minor}.{idAsVersion.Build}", buildResult.BuildVersionSimple);
         Assert.Equal(this.Repo.Head.Commits.First().Id.Sha, buildResult.GitCommitId);
+        Assert.Equal(this.Repo.Head.Commits.First().Author.When.UtcTicks.ToString(), buildResult.GitCommitDateTicks);
```
AArnott: All calls to `long.ToString` should be to the overload that can take a `CultureInfo.InvariantCulture` argument. This ensures it works reliably even in cultures that format numbers as strings in various ways.

src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs
```diff
@@ -150,6 +152,7 @@ private CodeTypeDeclaration CreateThisAssemblyClass()
                     { "AssemblyCompany", this.AssemblyCompany },
                     { "AssemblyConfiguration", this.AssemblyConfiguration },
                     { "GitCommitId", this.GitCommitId },
+                    { "GitCommitDate", this.GitCommitDateTicks },
```
AArnott: Can we make this field be declared as a `static readonly DateTimeOffset` instead of a `long`? It seems that would be much more readily useful to the consumer.
Same in the method below as well. It looks like below you *also* create a `DateTimeOffset` field. Why do we need the ticks version then?

AArnott:
We use CodeDom on .NET Framework to support as many languages as we can so long as a CodeDOM provider exists for it.
On .NET Core where those don't exist, we have to fallback to manual generators.
Due to the perf cost of CodeDOM, I actually think we recently switched to *prefer* the manual generators if we have them, and only fallback to CodeDOM on .NET Framework if we don't have a manual one. Is that not what's happening here?

src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs
qmfrederik: Yes, we can. I've updated the code.
Because CodeDOM doesn't have a way to create readonly fields, it becomes a read-only property instead.

qmfrederik:
I've updated the CodeDOM path, so hopefully we're good there. 

PS - the `Execute` method on `NET461` seems to always use CodeDOM (or error out in line 114 if CodeDOM doesn't support the current language)

PR #330, opened 2019-04-30
WIP: Add support for Git flow

### LINKED ISSUE #290: Support for Gitflow workflows
We use a [gitflow workflow](https://nvie.com/posts/a-successful-git-branching-model/) where the master branch is a release branch while a develop branch exists as a stabilization branch. Topic branches merge into the develop branch from which we release pre-releases for testing. Upon successful testing, we merge develop into master to create a release.

To date, NBGV has been mostly supporting this flow through my misunderstanding of "non-public releases" to mean "pre-releases". That is, with the default settings for cloud builds to include the commit ID, we do get something like "1.2.3+abcd1234" for topic and develop branches, while our master branch might build a 1.2.4 (we create a merge commit when merging from develop to release; otherwise, I realize this breaks semver). We then merge from master to develop (`--ff-only`) to pick up the merge commit and development continues with topic branches off develop.

The only thing not ideal with NBGV is that the build number uses build metadata (the default for cloud builds). We'd like to have NBGV set the prer-elease metadata instead (with or without the leading "g" - probably without given you recently dropped it from all but nupkg packages) to the pre-release format like it does for nupkgs.

Realizing this would be a breaking change, I propose opting into it by adding a third enum value to \#/cloudBuild/buildNumer/includeCommitId/where like "preReleaseForNonPublicReleases" or something suitably verbose (lacking the ability to document enum values in JSON schema).

All this would really change is instead of 1.2.3+abcd1234 our cloud build number would be 1.2.3-abcd1234. And in our case, this build number then easily carries through our build to release pipelines (Azure Pipelines) without having to persist across pipelines otherwise.

heaths:
Fixes #290 by adding new "where" property value "preReleaseForNonPublicRelease".

This is merely a preview of the changes per offline discussion.

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -917,7 +917,12 @@ public class BuildIntegrationTests : RepoTestBase
         Version assemblyVersion = GetExpectedAssemblyVersion(versionOptions, version);
         var additionalBuildMetadata = from item in buildResult.BuildResult.ProjectStateAfterBuild.GetItems("BuildMetadata")
                                       select item.EvaluatedInclude;
-        var expectedBuildMetadata = $"+{commitIdShort}";
+
+        var buildNumberOptions = versionOptions.CloudBuildOrDefault.BuildNumberOrDefault;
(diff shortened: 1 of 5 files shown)


heaths:
I've not had time to get back to this, and no longer have a need.

PR #333, opened 2019-05-14
Add support for BuildingTag in VisualStudioTeamServices

### LINKED ISSUE #332: Add support for BuildingTag in VisualStudioTeamServices
Hello,

after some investigation it seems like there might be some way to improve the VisualStudioTeamServices aka Azure DevOps integration.

Right now the `BuildingTag` is [not supported in Azure DevOps](https://github.com/AArnott/Nerdbank.GitVersioning/blob/63e0ff3130ed4e17225d6d56a21238cf4c81f2b6/src/NerdBank.GitVersioning/CloudBuildServices/VisualStudioTeamServices.cs#L18), but I think this can be improved and made to work.

I read [here](https://edsnider.net/2018/11/14/detecting-git-tag-pushes-in-azure-devops.html) that the `BUILD_SOURCEBRANCH` environment variable will start with `refs/tags/` when its building a tag.

So my idea is, when `BUILD_SOURCEBRANCH` starts with `refs/tags/` return a value from `BuildingTag` and return nothing from `BuildingBranch`.
Otherwise keep the existing behavior.

I'm just not sure in what kind format to return the value from `BuildingTag` - with or without the `refs/tags/` prefix?
In the AppVeyor integration, it seems to return only the tag name, without prefix.
But in the [version.json documentation](https://github.com/AArnott/Nerdbank.GitVersioning/blob/63e0ff3130ed4e17225d6d56a21238cf4c81f2b6/doc/versionJson.md) is a example with the prefix.

Any ideas or worries about this approach? 
Would gladly create a PR.

haefele:
This improves the VisualStudioTeamServices aka Azure DevOps integration, mostly just adds support for `BuildingTag`.

I removed one case from the BuildIntegrationTests, because in my tests even GitHub builds on Azure DevOps had the `refs/heads/` prefix.

Closes #332

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -450,7 +450,6 @@ public class BuildIntegrationTests : RepoTestBase
         {
             new object[] { CloudBuild.AppVeyor.SetItem("APPVEYOR_REPO_BRANCH", branchName) },
             new object[] { CloudBuild.VSTS.SetItem("BUILD_SOURCEBRANCH", $"refs/heads/{branchName}") },
-            new object[] { CloudBuild.VSTS.SetItem("BUILD_SOURCEBRANCH", branchName) }, // VSTS building a github repo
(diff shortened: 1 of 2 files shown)


AArnott:
Thanks!

PR #334, opened 2019-05-20
Exclude runtime assets for MSBuild

qmfrederik:
It took me a while to get `BuildIntegrationTests.AssemblyInfo` to work on my machine - it used to complain about two different versions of the MSBuild assemblies being loaded.

If I'm not mistaken, we can avoid this by adding `ExcludeAssets='runtime'` to the package reference for the MSBuild packages. That way, the assemblies which are sourced from NuGet are not copied to the output folder, and the MSBuild assemblies which ship with the .NET SDK would be used - pretty much the same way the MSBuildLocator works. https://docs.microsoft.com/en-us/visualstudio/msbuild/updating-an-existing-application?view=vs-2019#use-microsoftbuildlocator

It seemed to work locally, creating a PR to trigger CI (and feedback).

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/AssemblyInfoTest.cs
+++ b/src/NerdBank.GitVersioning.Tests/AssemblyInfoTest.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
-using Microsoft.Build.Locator;
 using Nerdbank.GitVersioning.Tasks;
 using Xunit;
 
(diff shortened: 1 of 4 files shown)


AArnott:
@qmfrederik I am reverting this change in #338 because it's breaking my local unit test debugging experience. If this can be done such that I can debug a test in the BuildIntegrationTests class using VS Test Explorer without a failure, then we can consider reapplying.

AArnott:
Never mind. I'm not going to revert it. Just refine it in #339.

PR #337, opened 2019-05-27
Fix the GitLab cloud provider, add unit tests

### LINKED ISSUE #335: Tags and GitLab CI
I want to build a public release when tagged using GitLab CI, and it seems that Nerdbank.GitVersioning is using the `CI_COMMIT_TAG` on GitLab.

Looking at the value of that variable on a tagged release, it is equal to the tag name without the `/refs/tags` prefix.

Should I be using `^v\\d+\\.\\d+` as my regex then on GitLab CI? Or is something not working as expected?

(Partially opening this issue just to help anyone who comes across the same problem)

qmfrederik:
This improves how the GitLab cloud provider handles refs and tags (although some level of guesswork is involved, as Gitlab doesn't provide raw access to the full ref).

Also adds unit tests. This requires mocking the `Environment.GetEnvironmentVariable` call, so there now is a `CloudBuildEnvironment` class with just a single `GetEnvironmentVariable` method which can be overriden for unit tests.

A bunch of classes are no longer `static` so we can use constructor DI on them.

Fixes #335

### DIFF
--- a/src/NerdBank.GitVersioning/CloudBuildServices/GitLab.cs
+++ b/src/NerdBank.GitVersioning/CloudBuildServices/GitLab.cs
@@ -17,7 +17,7 @@
             Environment.GetEnvironmentVariable("CI_COMMIT_TAG") == null ? 
                 $"refs/heads/{Environment.GetEnvironmentVariable("CI_COMMIT_REF_NAME")}" : null;
 
-        public string BuildingRef => this.BuildingBranch != null ? this.BuildingBranch : this.BuildingTag;
+        public string BuildingRef => this.BuildingBranch ?? this.BuildingTag;
 
         public string BuildingTag =>
             Environment.GetEnvironmentVariable("CI_COMMIT_TAG") != null ?
--- a/src/NerdBank.GitVersioning/CloudBuildServices/GitLab.cs
+++ b/src/NerdBank.GitVersioning/CloudBuildServices/GitLab.cs
@@ -2,8 +2,6 @@
 {
     using System;
     using System.Collections.Generic;
-    using System.ComponentModel;
-    using System.Diagnostics;
     using System.IO;
 
     /// <summary>
@@ -15,11 +13,15 @@
     /// </remarks>
     internal class GitLab : ICloudBuild
     {
-        public string BuildingBranch => Environment.GetEnvironmentVariable("CI_COMMIT_REF_NAME");
+        public string BuildingBranch =>
+            Environment.GetEnvironmentVariable("CI_COMMIT_TAG") == null ? 
+                $"refs/heads/{Environment.GetEnvironmentVariable("CI_COMMIT_REF_NAME")}" : null;
 
-        public string BuildingRef => Environment.GetEnvironmentVariable("CI_COMMIT_TAG");
+        public string BuildingRef => this.BuildingBranch != null ? this.BuildingBranch : this.BuildingTag;
 
-        public string BuildingTag => Environment.GetEnvironmentVariable("CI_COMMIT_TAG");
+        public string BuildingTag =>
+            Environment.GetEnvironmentVariable("CI_COMMIT_TAG") != null ?
+                $"refs/tags/{Environment.GetEnvironmentVariable("CI_COMMIT_TAG")}" : null;
 
         public string GitCommitId => Environment.GetEnvironmentVariable("CI_COMMIT_SHA");
 
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -48,6 +48,15 @@ public class BuildIntegrationTests : RepoTestBase, IClassFixture<MSBuildFixture>
     public BuildIntegrationTests(ITestOutputHelper logger)
         : base(logger)
     {
+        // MSBuildExtensions.LoadMSBuild will be called as part of the base constructor, because this class
+        // implements the IClassFixture<MSBuildFixture> interface. LoadMSBuild will load the MSBuild assemblies.
+        // This must happen _before_ any method that directly references types in the Microsoft.Build namespace has been called.
+        // Net, don't init MSBuild-related fields in the constructor, but in a method that is called by the constructor.
+        this.Init();
+    }
+
+    private void Init()
+    {
         int seed = (int)DateTime.Now.Ticks;
         this.random = new Random(seed);
         this.Logger.WriteLine("Random seed: {0}", seed);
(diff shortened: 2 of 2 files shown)


src/NerdBank.GitVersioning/CloudBuildServices/GitLab.cs
```diff
@@ -15,15 +14,26 @@
...
 
-        public string BuildingRef => Environment.GetEnvironmentVariable("CI_COMMIT_TAG");
+        public GitLab(CloudBuildEnvironment buildEnvironment)
+        {
+            this.buildEnvironment = buildEnvironment ?? throw new ArgumentNullException(nameof(buildEnvironment));
+        }
+
+        public string BuildingBranch =>
+            this.buildEnvironment.GetEnvironmentVariable("CI_COMMIT_TAG") == null ? 
+                $"/refs/heads/{this.buildEnvironment.GetEnvironmentVariable("CI_COMMIT_REF_NAME")}" : null;
```
AArnott: There should be no leading `/` in the return value

src/NerdBank.GitVersioning.Tests/CloudBuildServices/GitLabTests.cs
```diff
@@ -0,0 +1,53 @@
...
+                    new Dictionary<string, string>()
+                    {
+                        { "CI_COMMIT_REF_NAME", "master" },
+                        { "CI_COMMIT_SHA", "1ecfd275763eff1d6b4844ea3168962458c9f27a" },
+                        { "GITLAB_CI", "true" }
+                    }));
+
+
+            Assert.NotNull(cloudBuild.Active);
+            Assert.Equal("/refs/heads/master", cloudBuild.Active.BuildingBranch);
```
AArnott: Update this assertion as well to *not* expect a leading slash.

src/NerdBank.GitVersioning/CloudBuildEnvironment.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// Provides access to the current system's environment variables.
+    /// </summary>
+    public class CloudBuildEnvironment
+    {
+        /// <summary>
+        /// Gets the default instance of the <see cref="CloudBuildEnvironment"/> class.
+        /// </summary>
+        public static CloudBuildEnvironment Instance
+        { get; set; } = new CloudBuildEnvironment();
```
AArnott: Move this up to the end of the previous line, please.

src/NerdBank.GitVersioning/CloudBuildEnvironment.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+
+namespace NerdBank.GitVersioning
+{
+    /// <summary>
+    /// Provides access to the current system's environment variables.
+    /// </summary>
+    public class CloudBuildEnvironment
```
AArnott: This class should be internal. It's not part of the public contract I wish to maintain with folks who reference this library.

qmfrederik:
@AArnott Sure, I updated the branch to use the existing pattern, and removed the leading /.

I had to update the constructor of the `BuildIntegrationTests` to not directly use any MSBuild-related types; otherwise the unit tests would fail with this error:

```
System.InvalidOperationException : Microsoft.Build.Locator.MSBuildLocator.RegisterInstance was called, but MSBuild assemblies were already loaded.
Ensure that RegisterInstance is called before any method that directly references types in the Microsoft.Build namespace has been called.
Loaded MSBuild assemblies: Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
Microsoft.Build.Framework, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
Stack Trace:
   at Microsoft.Build.Locator.MSBuildLocator.RegisterMSBuildPath(String msbuildPath)
   at Microsoft.Build.Locator.MSBuildLocator.RegisterInstance(VisualStudioInstance instance)
   at Microsoft.Build.Locator.MSBuildLocator.RegisterDefaults()
   at MSBuildExtensions.LoadMSBuild()
   at BuildIntegrationTests..ctor(ITestOutputHelper logger)
```

Let me know if this works for you.

AArnott:
> I updated the branch to use the existing pattern, and removed the leading /.

Thanks

> I had to update the constructor of the BuildIntegrationTests to not directly use any MSBuild-related types; otherwise the unit tests would fail with this error:

I already fixed it [here](https://github.com/AArnott/Nerdbank.GitVersioning/pull/339/files#diff-671f15e0b30fa54620eef736ecea8be5). If you revert your own fix for it and merge from my `v2.3` branch, you'll get the fix too.

Note that over the weekend I bumped `master` to `v3.0-beta`, so I'll retarget you PR to the `v2.3` branch since that's effectively what you were targeting before.

AArnott:
Never mind. I can't retarget it because you've already rebased onto the new master branch. May I recommend you rebase your commits onto v2.3 and change the target branch of this PR yourself when ready?

src/NerdBank.GitVersioning/CloudBuildServices/GitLab.cs
```diff
@@ -15,11 +13,15 @@
     /// </remarks>
     internal class GitLab : ICloudBuild
     {
-        public string BuildingBranch => Environment.GetEnvironmentVariable("CI_COMMIT_REF_NAME");
+        public string BuildingBranch =>
+            Environment.GetEnvironmentVariable("CI_COMMIT_TAG") == null ? 
+                $"refs/heads/{Environment.GetEnvironmentVariable("CI_COMMIT_REF_NAME")}" : null;
 
-        public string BuildingRef => Environment.GetEnvironmentVariable("CI_COMMIT_TAG");
+        public string BuildingRef => this.BuildingBranch != null ? this.BuildingBranch : this.BuildingTag;
```
AArnott:
```suggestion
        public string BuildingRef => this.BuildingBranch ?? this.BuildingTag;
```

qmfrederik:
>> I had to update the constructor of the BuildIntegrationTests to not directly use any MSBuild-related types; otherwise the unit tests would fail with this error:

> I already fixed it here. If you revert your own fix for it and merge from my v2.3 branch, you'll get the fix too.

That fix doesn't seem to be sufficient, and I had to move all code which references any MSBuild-related type out of the constructor.

> May I recommend you rebase your commits onto v2.3 and change the target branch of this PR yourself when ready?

Sure, done.

AArnott:
> That fix doesn't seem to be sufficient, and I had to move all code which references any MSBuild-related type out of the constructor.

Strange. But oh well. Thanks for contributing!

PR #338, opened 2019-05-27
Revert MSBuildLocator change

AArnott:


AArnott:
Hmmm... I'm not sure this fixes it either. With the revert, the local test debugging experience still fails, but with this error:

> 7: The "Message" task could not be instantiated from the assembly "Microsoft.Build.Tasks, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a". Please verify the task assembly has been built using the same version of the Microsoft.Build.Framework assembly as the one installed on your computer and that your host application is not missing a binding redirect for Microsoft.Build.Framework. Unable to cast object of type 'Microsoft.Build.Tasks.Message' to type 'Microsoft.Build.Framework.ITask'.

AArnott:
Wow, this really is a mess. The MSBuild assemblies that are referenced by the product are all v15.1, yet in the test environment MSBuild.Framework is loaded twice, and the tasks/utilities assemblies that load are 2.0 assemblies. I've definitely got cleanup to do.

PR #345, opened 2019-05-29
Add doc that explains public VS. Stable releases 

AArnott:


doc/public_vs_stable.md
```diff
@@ -0,0 +1,122 @@
...
+From semver.org:
+
+> A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version.
+
+The unstable nature of a product might be in functional resilience, or that its API isn't finalized, or lack of adequate testing.
+Any and all of these are based on the assessment of the software engineers responsible for the project.
+
+Like the version *number*, the `-prerelease` tag (if there is one) is recorded in a git source tree for Nerdbank.GitVersioning to use when building.
+A given commit in a repo represents software that builds v1.2 of a product or v1.2-beta of a product, depending on how its owner(s) felt about the commit at the time they authored it.
+When a branch becomes stable, the `-prerelease` tag can be removed by adding a commit to the branch that strips the tag.
```
MarcoRossignoli: new line and line space after this...phrase after is the key to me...maybe should be bold

MarcoRossignoli:
@AArnott out of scope...I did't take a look at code but you can answer quickly...to count height do you check every update of `version.json` file and compare schemas querying repo history?

doc/public_vs_stable.md
```diff
@@ -0,0 +1,122 @@
...
+Folks sometimes ask if there is a way to remove the `-prerelease` tag from an existing commit that has a `-prerelease` tag expressed in its committed version.json. There is not.
+To remove the `-prerelease`, the version.json file must be changed to remove it.
+Committing this change communicates to everyone looking at the repo that this software is stable.
+
+The natural evolution of a product includes usually includes entering and exiting a `-prerelease` stage many times, but within a branded release (usually recognized by an intentional version number like "1.2") the progression usually transitions only one direction: from `-prerelease` to stable quality.
```
MarcoRossignoli: So could be possible have a master always in `-beta` and when someone wants release remove beta, do commit generate package and relese and right after re-add `-beta` to start a "new pahase"?

AArnott:
@MarcoRossignoli  To count height, I walk history by commit from HEAD to the last commit that retains the same `version` property *integers*. So changing major.minor or major.minor.patch will reset the height, but adding or removing a `-beta` will not reset height.
Then there's logic to deal with branches with multiple ways to get to the last commit. This I believe is documented in the README.

doc/public_vs_stable.md
AArnott: Done.

doc/public_vs_stable.md
AArnott: possible, yes, but tedious. It doesn't leave you a branch for servicing that release either.
For this (and I'll add this to the doc), we offer the [`nbgv prepare-release`](https://github.com/AArnott/Nerdbank.GitVersioning/blob/master/doc/nbgv-cli.md#preparing-a-release) command.

MarcoRossignoli:
> This I believe is documented in the README.

Yes I wonder only how it's implemented, you walk commit history for version.json file...maybe there was some other trick or you booked ver in some git metadata place, but it's not so.

doc/public_vs_stable.md
MarcoRossignoli: Great clear.

AArnott:
Version height calculation [starts here](https://github.com/AArnott/Nerdbank.GitVersioning/blob/18ec2bd6b831fa787bfbe8664eecf08953f0e2f9/src/NerdBank.GitVersioning/GitExtensions.cs#L65).

doc/public_vs_stable.md
MarcoRossignoli: @AArnott many compliments...ngbv lib+commands are a bomb for versioning lib...I'try to apply "all concepts" step by step with @tonerdo

MarcoRossignoli:
@AArnott in a case like our where we have 3 components with different version for historical reason we should update version.json like

release.branchName = "release_msbuild/v{version}"
release.branchName = "release_collector/v{version}"
release.branchName = "release_console/v{version}"

and run nbgv `prepare-release` inside project folder, right?

AArnott:
@MarcoRossignoli ugh. I don't know how well that will work or if there's a better way. I've never heard of quite your situation before when using `prepare-release`. I suppose you could try it and see.
But if you plan to release all at the same time as a rule, personally I might try for maintaining just one branch for each release instead of 3. That might mean that `prepare-release` isn't useful.
On the other hand if you might release just one package for servicing, then one branch per package might make sense.

MarcoRossignoli:
> On the other hand if you might release just one package for servicing, then one branch per package might make sense.

Eh at the moment is our use case, because the "core" is shared between 3 "drivers" but not all support all code feature(passing parameters) and "drivers" doesn't provide same "added features", but this is a "design" issue to me, maybe in future we will be able to "merge features" in core and let "drivers" to be only "pluggable entry points".

I agree...if we'll decide to "align" version some day maybe will be better.

BTW I did some tests on my custom repo, tool and the idea works very well.

doc/public_vs_stable.md
```diff
@@ -0,0 +1,127 @@
...
+For example, an anticipated version 1.2 might first be released to the public as 1.2-beta before releasing as 1.2 (without the `-beta` suffix).
+If the product is undergoing significant changes that warrant downgrading the stability rating to pre-release quality, the version number tends to be incremented at the same time.
+So a 1.2 product's subsequent release might appear as 1.3-beta or 2.0-beta.
+But for a particularly stable product, it's possible for releases to remain stable from one release (1.2) to the next (1.3) without ever publishing a pre-release version.
+
+**Tip**: To aid in the common workflow of stabilizing for a release including branching and updating `version.json`, and mitigating merge conflicts in that file, we have the [`nbgv prepare-release`][nbgv_prepare-release] command to automate the process.
+
+In all this, to consumers of the product there is never any question regarding which of two releases is newer.
+[SemVer formalizes version comparisons](https://semver.org/#spec-item-11) but in essence, the larger the number, the newer it is such that there is never ambiguity between two versions.
```
heaths: "... but, in essence, ..."

heaths: "... the larger the number the newer it is ..." (no comma needed).

doc/public_vs_stable.md
```diff
@@ -0,0 +1,127 @@
...
+  ]
+}
+```
+
+When you specify multiple branches as public release branches, it is very important that each of these branches have a *unique* version specified in the `version` property of the version.json file.
+This guarantees that versions built from any two of these public release branches never collide in version number.
+Naming most/all your public release branches after the version they build can help folks to find the right branch as well as help maintain unique versions for each branch.
+
+In development of a topic branch, you might find a need to share packages before merging into one of these public release branches.
+That's just fine--you can share your `-gc0ffeebeef` suffixed packages.
```
heaths: You have spaces surrounding the rendered em dash further down. May want them here as well.

doc/public_vs_stable.md
AArnott: Fixed

doc/public_vs_stable.md
AArnott: Fixed

PR #359, opened 2019-06-12
Fix VB code generator

### LINKED ISSUE #358: Version 2.3.179 creates VB syntax error in ThisAssembly class
The new version 2.3.179 creates invalid code for Visual Basic projects for the `ThisAssembly` class.

The produced code is the following:
```VB.NET
<Assembly: System.Reflection.AssemblyVersionAttribute("…")>
<Assembly: System.Reflection.AssemblyFileVersionAttribute("…")>
<Assembly: System.Reflection.AssemblyInformationalVersionAttribute("…")>
#If NETSTANDARD Or NETFRAMEWORK Or NETCOREAPP Then
<System.CodeDom.Compiler.GeneratedCode("Nerdbank.GitVersioning.Tasks","2.3.179.31651")>
#End If
Partial Friend NotInheritable Class ThisAssembly
    …
End Class
```

This is not valid in Visual Basic, because it's illegal for anything but more attributes to be written between the declaration of a type (or a member for that matter) and the attribute that will be applied to it. This includes the `#End If` instruction.

The only way to build what you did here as valid code is to include the declaration of the `class` in the header. Like so:
```VB.NET
<Assembly: System.Reflection.AssemblyVersionAttribute("…")>
<Assembly: System.Reflection.AssemblyFileVersionAttribute("…")>
<Assembly: System.Reflection.AssemblyInformationalVersionAttribute("…")>
#If NETSTANDARD Or NETFRAMEWORK Or NETCOREAPP Then
<System.CodeDom.Compiler.GeneratedCode("Nerdbank.GitVersioning.Tasks","2.3.179.31651")>
Partial Friend NotInheritable Class ThisAssembly
#Else
Partial Friend NotInheritable Class ThisAssembly
#End If
    …
End Class
```

mkaring:
This is the pull request to fix #358

### DIFF
--- a/src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs
+++ b/src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs
@@ -541,8 +541,10 @@
             {
                 this.codeBuilder.AppendLine($"#If {CompilerDefinesAroundGeneratedCodeAttribute.Replace("||", " Or ")} Then");
                 this.codeBuilder.AppendLine($"<System.CodeDom.Compiler.GeneratedCode(\"{GeneratorName}\",\"{GeneratorVersion}\")>");
-                this.codeBuilder.AppendLine("#End If");
                 this.codeBuilder.AppendLine("Partial Friend NotInheritable Class ThisAssembly");
+                this.codeBuilder.AppendLine("#Else");
+                this.codeBuilder.AppendLine("Partial Friend NotInheritable Class ThisAssembly");
+                this.codeBuilder.AppendLine("#End If");
             }
 
             internal override void AddThisAssemblyMember(string name, string value)


AArnott:
Hmmm... this looks like a regression in the VB compiler. The build integration tests for VB are clearly running _and passing_ yet I can see in the latest VS 2019 builds same as you, that it introduces a compiler error.

mkaring:
I am using the latest VS2017 version (15.9.21) and I see the error. It also is correct that the error is shown, because it is a syntax error. According to the [documentation][1] the attributes have to be "on the same line" as the declaration. Breaking the line only works because it utilizes the [implicit line continuation][2]. The preprocessor statements interrupt the implicit continuation and so the error is correct.

[1]: https://docs.microsoft.com/dotnet/visual-basic/programming-guide/concepts/attributes/#using-attributes
[2]: https://docs.microsoft.com/de-de/dotnet/visual-basic/programming-guide/language-features/statements#implicit-line-continuation

AArnott:
It's definitely a vb compiler or .NET SDK issue. It only repros for .NET SDK style projects.
I filed https://github.com/dotnet/roslyn/issues/36365

In the meantime though, of course we want NB.GV to work for VB .NET SDK style projects, so we'll take the PR. :)

PR #381, opened 2019-09-04
Add fsharp and vb integration tests

### LINKED ISSUE #353: Add F# build integration test
We need an F# build integration test to avoid bugs like #349

pihai:
This adds an integration test for F# (#353) and also for VB.

In F# the following warning is emitted: `warning FS2003: The attribute System.Reflection.AssemblyInformationalVersionAttribute specified version '42.42.1+2146ddd457', but this value is invalid and has been ignored [/mnt/c/temp/nbgv-int-test/lib/lib.fsproj]`

I noticed that the F# test fails with dotnet SDK 2.2. It wants to download the FSharp.Core package from nuget, but the official feed is removed in the created nuget.config:
error: Unable to find package FSharp.Core. No packages exist with this id in source(s): local
But with 2.1 it should be fine.

Closes #353

### DIFF
--- a/azure-pipelines/xplattest-pipeline.yml
+++ b/azure-pipelines/xplattest-pipeline.yml
@@ -31,7 +31,7 @@ steps:
       echo Testing language $l &&
       dotnet new classlib -lang=$l &&
       dotnet add package nerdbank.gitversioning -v $NBGV_NuGetPackageVersion &&
-      dotnet pack -c debug &&
+      dotnet pack -c debug /p:TreatWarningsAsErrors=false &&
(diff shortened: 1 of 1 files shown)


pihai:
Hm it seems that the warning mentioned in the previous comment is an error on Mac:
/Users/vsts/agent/2.155.1/work/1/s/lib/obj/debug/netstandard2.0/lib.Version.fs(14,69): error FS2003: The attribute System.Reflection.AssemblyInformationalVersionAttribute specified version '42.42.1+bfae5dcc30', but this value is invalid and has been ignored [/Users/vsts/agent/2.155.1/work/1/s/lib/lib.fsproj]

PR #394, opened 2019-10-15
Add support for worktrees

### LINKED ISSUE #152: Fails within a git 'worktree'
The experimental [git worktree](https://git-scm.com/docs/git-worktree) feature creates a checked out directory that doesn't have a .git folder, but rather a .git file that points to a folder within the .git folder in another directory. NB.GV should find that and work with it, if libgit2sharp supports it.

AArnott:
Fixes #152


AArnott:
Worktrees (accidentally) started working in the v3.0.48 release already.

PR #399, opened 2019-10-23
Add path filtering support to version.json

### LINKED ISSUE #160: All packages in repo get version bump for every commit, even irrelevant ones

saul:
Goes part of the way to solving #160 (superseding #223 which has been quiet for most of the year).

Opening this now in a WIP state to get early feedback.

This PR adds a new `pathFilters` property to `version.json`. The array defines the paths to include in version height calculations. If a commit does not affect any path in `pathFilters`, it does not increment the version height. This feature is entirely opt-in - if `pathFilters` is not defined, NB.GV calculates version height as it previously did.

**Todo**:

- [x] Update JSON schema
- [x] Update existing tests
- [x] Write new tests
- [x] Write monorepo guidance explaining how to use this feature

Closes #160


src/NerdBank.GitVersioning/version.schema.json
```diff
@@ -175,6 +175,14 @@
             }
           },
           "additionalProperties": false
+        },
+        "pathFilters": {
+          "type": "array",
```
AArnott: I feel like we may want both _include_ paths and _exclude_ paths, where the excluded paths would presumably be subsets of include paths.
We could do this with `include` and `exclude` properties under `pathFilters`, or we could do a straight string array where order is significant and exclude paths start with `!`.
Allowing relative paths here sounds likely useful (although I imagine most of them would be relative with `../` leading paths, wouldn't they? What kind of paths would you expect for these repos?
I think supporting repo-root relative paths is important as well. In git the syntax for this is a leading `:/` string, which we should consider adopting here.

src/NerdBank.GitVersioning/version.schema.json
```diff
@@ -175,6 +175,14 @@
...
+          "description": "An array of repository relative paths (folders or files) that are used to filter commits when calculating the version height. A commit will not increment the version height if it does not touch a path listed in this array.",
+          "items": {
+            "type": "string"
```
AArnott: I wonder if there's some kind of path-compatible regex we can apply here in the schema. Perhaps to help encourage folks to use only forward slashes, or start with a leading slash, or anything else that we'd need to process these paths efficiently.

src/NerdBank.GitVersioning/version.schema.json
saul: `pathFilters` is passed verbatim to LibGit2Sharp.Diff.Compare, which in turn passes it through to libgit2. These paths don't actually only support file and folders (as the doc comment in LibGit2Sharp suggests), but they are pathspecs. This means the user is free to do things like `:^DirectoryToExclude/` to exclude paths.

It doesn't appear to be possible to make the paths relative to the project directory, without mapping the paths ourselves in NB.GV such that they are relative to the root. Maybe we should do the following:

- Any paths starting with `:` are treated as pathspecs and passed to Diff.Compare verbatim
- Relative paths are combined with the repo root to produce repo root relative paths

src/NerdBank.GitVersioning/version.schema.json
saul: Done in the latest commit

saul:
@AArnott mind giving this another review? I've added some tests now and I'm happy that it's working as expected.

shana:
I was testing out the latest build from this PR on one of my projects, and as soon as I use bump the package version to the one from this PR, the generated version stops incrementing (it's always 1.0.0), so something here breaks the default behaviour. I can't get it to increment with pathFilters, either.

saul:
@AArnott this is ready for review when you get a moment.

src/NerdBank.GitVersioning/FilterPath.cs
```diff
@@ -0,0 +1,155 @@
...
+                }
+            }
+            else
+            {
+                this.RepoRelativePath = ParsePath(pathSpec, relativeTo);
+            }
+
+            this.RepoRelativePath =
+                this.RepoRelativePath
+                    .Replace('\\', '/')
```
Smaug123: I take it we don't care about any Linux users who were so bold as to use backslashes in their folder names?

saul: I guess not 😄 For what it's worth, files with backslashes committed within a Git repo are impossible to checkout on Windows anyway.

src/NerdBank.GitVersioning/FilterPath.cs
```diff
@@ -0,0 +1,155 @@
...
+using System.Linq;
+using LibGit2Sharp;
+using Validation;
+
+namespace Nerdbank.GitVersioning
+{
+    /// <summary>
+    /// A filter (include or exclude) representing a repo relative path.
+    /// </summary>
+    public class FilterPath
```
AArnott: Can we keep this class internal so it doesn't increase our public API surface area that we need to support? Is there any reason it needs to be exposed?

src/NerdBank.GitVersioning/FilterPath.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// </summary>
+        public bool IsExclude { get; }
+
+        /// <summary>
+        /// Path relative to the repository root that this <see cref="FilterPath"/> represents.
+        /// Slashes are canonical for this OS.
+        /// </summary>
+        public string RepoRelativePath { get; }
+
+        private static string ParsePath(string path, string relativeTo)
```
AArnott: This method lacks an xml doc comment explaining what it does, and in 10+ seconds of reading it I can't figure it out either. Please add xml doc comments.

src/NerdBank.GitVersioning/FilterPath.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// <summary>
+        /// Calculate the <see cref="FilterPath"/>s for a given project within a repository.
+        /// </summary>
+        /// <param name="versionOptions">Version options for the project.</param>
+        /// <param name="relativeRepoProjectDirectory">
+        /// Path to the project directory, relative to the root of the repository.
+        /// If <c>null</c>, assumes root of repository.
+        /// </param>
+        /// <param name="repository">Git repository containing the project.</param>
+        /// <returns>A list of <see cref="FilterPath"/> instances.</returns>
```
AArnott: apparently `null` is also a possible result. Please document that here.

src/NerdBank.GitVersioning/FilterPath.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// <returns>
+        /// True if this <see cref="FilterPath"/> is an excluding filter that matches
+        /// <paramref name="repoRelativePath"/>, otherwise false.
+        /// </returns>
+        public bool Excludes(string repoRelativePath)
+        {
+            if (!this.IsExclude) return false;
+
+            return this.RepoRelativePath.Equals(repoRelativePath, this.stringComparison) ||
+                   repoRelativePath.StartsWith(this.RepoRelativePath + "/",
```
AArnott: This will throw `NullReferenceException` if `repoRelativePath == null`. Can you mitigate that?

src/NerdBank.GitVersioning/GitExtensions.cs
```diff
@@ -116,34 +117,45 @@ public static int GetVersionHeight(this Branch branch, string repoRelativeProjec
...
+        /// <param name="pathFilters">Paths to include/exclude from height calculation.</param>
         /// <param name="continueStepping">
         /// A function that returns <c>false</c> when we reach a commit that
         /// should not be included in the height calculation.
         /// May be null to count the height to the original commit.
         /// </param>
         /// <returns>The height of the commit. Always a positive integer.</returns>
-        public static int GetHeight(this Commit commit, Func<Commit, bool> continueStepping = null)
+        public static int GetHeight(this Commit commit, IReadOnlyList<FilterPath> pathFilters,
+            Func<Commit, bool> continueStepping = null)
```
AArnott: stylecop: please don't spread parameters across lines unless each one gets its own line. And IMO, this signature is short enough to fit on one line.
(aside: I really need to add the stylecop.analyzers to this project)

src/NerdBank.GitVersioning/GitExtensions.cs
```diff
@@ -116,34 +117,45 @@ public static int GetVersionHeight(this Branch branch, string repoRelativeProjec
...
         {
             Requires.NotNull(commit, nameof(commit));
 
+            var includePaths =
+                pathFilters
+                    ?.Where(filter => !filter.IsExclude)
+                    .Select(filter => filter.RepoRelativePath)
+                    .ToList();
+
+            var excludePaths = pathFilters?.Where(filter => filter.IsExclude).ToList();
```
AArnott: Is it accurate to split out includes and excludes as if their intermingled order doesn't matter?

src/NerdBank.GitVersioning/GitExtensions.cs
```diff
@@ -263,7 +285,7 @@ public static Version GetIdAsVersion(this Repository repo, string repoRelativePr
         /// Looks up the commit that matches a specified version number.
         /// </summary>
         /// <param name="repo">The repository to search for a matching commit.</param>
-        /// <param name="version">The version previously obtained from <see cref="GetIdAsVersion(Commit, string, int?)"/>.</param>
+        /// <param name="version">The version previously obtained from <see cref="GetIdAsVersion(LibGit2Sharp.Commit,string,System.Nullable{int})"/>.</param>
```
AArnott: You have a few of these types of changes. Why? Can you revert them?

src/NerdBank.GitVersioning/GitExtensions.cs
```diff
@@ -116,34 +117,45 @@ public static int GetVersionHeight(this Branch branch, string repoRelativeProjec
...
         /// <param name="branch">The branch to measure the height of.</param>
+        /// <param name="pathFilters">Paths to include/exclude from height calculation.</param>
         /// <param name="continueStepping">
         /// A function that returns <c>false</c> when we reach a commit that
         /// should not be included in the height calculation.
         /// May be null to count the height to the original commit.
         /// </param>
         /// <returns>The height of the branch.</returns>
-        public static int GetHeight(this Branch branch, Func<Commit, bool> continueStepping = null)
+        public static int GetHeight(this Branch branch, IReadOnlyList<FilterPath> pathFilters, Func<Commit, bool> continueStepping = null)
```
AArnott: Changing the signature of a public method is a breaking API change (as attested to in fact by the no-op changes you had to make to some of the tests). Can you please add new methods that have the original signature and have them call into the new ones to avoid the breaking change?

src/NerdBank.GitVersioning/version.schema.json
```diff
@@ -175,6 +175,15 @@
             }
           },
           "additionalProperties": false
+        },
+        "pathFilters": {
+          "type": "array",
+          "description": "An array of pathspec-like strings that are used to filter commits when calculating the version height. A commit will not increment the version height if its changed files are not included by these filters.\nPaths are relative to this file. Paths relative to the root of the repository can be specified with the `:/` prefix.\nExclusions can be specified with a `:^` prefix for relative paths, or a `:^/` prefix for absolute paths.",
```
AArnott: Really good description. Thanks. But why do exclusions start with `:^` instead of just `!` ? The .gitignore syntax uses just a leading `!` character. And if `:/` is git's way of referring to the repo root, having the `^` character stick in between like `:^/` for repo-relative paths doesn't seem to match any precedent I can think of.
Also: let's refer in this description to all `:/`-like sequences not as "absolute paths" but as "repo-relative paths".

src/NerdBank.GitVersioning/GitExtensions.cs
```diff
@@ -58,7 +58,8 @@ public static int GetVersionHeight(this Commit commit, string repoRelativeProjec
             var versionHeightPosition = versionOptions.VersionHeightPosition;
             if (versionHeightPosition.HasValue)
             {
-                int height = commit.GetHeight(c => CommitMatchesVersion(c, baseSemVer, versionHeightPosition.Value, repoRelativeProjectDirectory));
+                var pathFilters = FilterPath.FromVersionOptions(versionOptions, repoRelativeProjectDirectory, commit.GetRepository());
```
AArnott: Why is it accurate/desirable to only consider the filters as they are set at this particular commit? If filters change over time, shouldn't we honor those as we walk back in history?

src/NerdBank.GitVersioning.Tests/GitExtensionsTests.cs
```diff
@@ -121,6 +123,74 @@ public void GetVersionHeight_VersionJsonHasParsingErrorsInHistory()
         Assert.Equal(0, this.Repo.GetVersionHeight());
     }
 
+    [Fact]
```
AArnott: Please add tests to confirm that:

* [ ] our `GitExtensions.GetCommitsFromVersion` method still works. One example: that it can find each commit before, within and after a span of commits that had the same version.
* [ ] *changes* to filters has the right behavior. The behavior *I* would expect is that each commit contributes to height based on the version.json files within that commit. If I add an exclude path to version.json, that shouldn't *lower* the version height after I check it in, but it would have that effect if you're only reading the filters at the originating commit. It would be quite bizarre, IMO.

src/NerdBank.GitVersioning/GitExtensions.cs
saul: Apologies - Rider appears to have done this. I'm happy to revert them - I assume the Rider version is more correct?

AArnott: No. The Rider version is equivalent, but more wordy than necessary, making it harder to read.

shana:
I found that if I set a path filter to the current directory as `"./"`, it won't properly update the version. Same for `"../[name of directory]". It really looks like relative paths to directories is not working.

The only way to monitor the current directory is to set its absolute path, which is kind of annoying, since it's nice to have a set of path filters that don't have hardcoded directory names that I can copy around to all the separate projects in the repo, like:

`"pathFilters": ["./", "../Directory.Build.props", "/icon.png", "/LICENSE.md", ":/common"],`

src/NerdBank.GitVersioning/version.schema.json
saul: I've tried to match Git's official definition of a 'pathspec' as closely as possible. See the documentation here: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec

> **exclude**
> After a path matches any non-exclude pathspec, it will be run through all exclude pathspecs (magic signature: ! or its synonym ^). If it matches, the path is ignored. When there is no non-exclude pathspec, the exclusion is applied to the result set as if invoked without any pathspec.

src/NerdBank.GitVersioning/GitExtensions.cs
saul: Yes - their relative order is irrelevant. I've tried to match Git's official definition of a 'pathspec' as closely as possible. See the documentation here: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec

> **exclude**
> After a path matches any non-exclude pathspec, it will be run through all exclude pathspecs (magic signature: ! or its synonym ^). If it matches, the path is ignored. When there is no non-exclude pathspec, the exclusion is applied to the result set as if invoked without any pathspec.

I'll include this info in the description in the JSON schema.

src/NerdBank.GitVersioning/FilterPath.cs
saul: We could keep it internal but then none of the methods on GitExtensions would be able to filter paths, which would be a shame. I created this class as an argument of `IReadOnlyList<FilterPath>` is far more obvious than `IEnumerable<string>` to represent filters.

In GitExtensions we could replace instances of `IReadOnlyList<FilterPath> filters` with `IEnumerable<string> include, IEnumerable<string> exclude` and then make this class internal. However I think having `FilterPath` on the public API makes it obvious.

Which would you prefer?

src/NerdBank.GitVersioning/GitExtensions.cs
saul: Good point. I've changed the logic to do that instead. One consideration is that if the project directory disappears as we walk up the history, we will instead treat that as no filters being present (i.e. assume all paths affect commit height). I think this is what we want - I'll write tests codifying this assumption.

saul:
@shana I find that surprising. There is already a test `GetVersionHeight_IncludeFilter` which uses a relative path to include the project directory, and it passes.

Can you confirm that you're definitely seeing this problem as you describe?

src/NerdBank.GitVersioning.Tests/GitExtensionsTests.cs
saul: Added test `GetCommitsFromVersion_WithPathFilters` for your first point. Is this what you were looking for?

As for the second point, added `GetVersionHeight_AddingExcludeDoesNotLowerHeight` to cover that.

I've also added a few more tests covering some edge cases. Let me know if that's suffice.

saul:
Hi @AArnott - sorry for taking so long to get back to you.

I think I've covered all of your review comments - I've left a couple of them unresolved as I'm still waiting for a response from you on that.

When you get a moment I think this is ready for re-review, thanks.

src/NerdBank.GitVersioning/FilterPath.cs
AArnott: I pity the fool
... who uses backslashes in their file names. I can't imagine how many things break.

src/NerdBank.GitVersioning/FilterPath.cs
AArnott: I like having `FilterPath` used in the extension method instead of just string. But is there any reason the extension method has to be public? Unless there's a known use case to expose an API, keeping it internal keeps options open later.

shana:
@saul Now I can't reproduce it anymore. It's really confusing. It might be some cached state in the dotnet build server itself - build tasks dlls don't get unloaded between builds, and if one project is using one version of a build task and another project another version, it causes all kinds of horrible build failures. I'm having to shutdown the build-server every time I switch projects, so maybe that's what's happening... 😕

AArnott:
Just FYI I'm hoping to review this in the next few days. My vacation has started so I'm catching up on OSS repo work.

src/NerdBank.GitVersioning/FilterPath.cs
AArnott: It appears nothing prevents `FilterPath` from being made internal. I simply compile it into the test project as well so that we can keep your unit tests.

src/NerdBank.GitVersioning.Tests/GitExtensionsTests.cs
AArnott: Yes. The tests look great. Thanks.

AArnott:
OK, these changes look good. Thanks for thoroughly addressing my concern about respecting filters at each individual commit.

My only remaining concern is perf of the `GetCommitHeight` method, which now does  *lot* more work per commit and allocates lots more memory (and thus produces GC pressure that can hurt overall throughput too). I think we can likely mitigate my concerns through caching the `VersionOptions` and `PathFilters` based on the blob sha of version.json so that we only pay the tax on commits that actually changed the file.  
Also the `continueStepping` delegate already calls `VersionFile.GetVersion` and your code adds yet another call to it. So to minimize perf impact (particularly to users who are not even using filters) I think we need to apply this cache such that it can be shared with that `continueStepping` delegate as well.

Do you want to try resolving this? If not, I will try to find time over the next couple weeks.

saul:
Thanks for looking at this again @AArnott. I'm away from a dev machine for the holidays so can't contribute in a meaningful way until January. I'm happy to resolve the perf concerns when I return (unless you have some free time before then).

Another thing that crossed my mind which needs resolving (and testing) is invalid `pathFilters` in the `version.json` file. We need to be resilient to that. At the moment I think an exception would cause the height methods to crash.

AArnott:
I don't think I'm too concerned about invalid paths. If the json can't parse we throw as well. People need to edit any commits that introduce a failure, so they are unlikely to ever push such commits in the first place.

saul:
@AArnott I've implemented a partially working cache, but I'm having some problems. The cache (as written) assumes that the object returned by `VersionFile.GetVersion` is not mutated. If it's mutated, then the cached `VersionOptions` are clobbered. This is the reason that those 2 tests are failing in CI - version.json inheritance causes the object returned by `GetVersion` to be mutated.

With this in mind there are a couple of options:

1. If we find a cached version.json in `VersionFile.GetVersion`, we clone it before returning. This leaves the caller to mutate the object as they wish. We can implement the cloning by:
  a. Implementing ICloneable/MemberwiseClone on `VersionFile`. This is brittle as it must be manually updated when new properties are added to `VersionFile`.
  b. Serialise the VersionFile, then deserialise it. This completely defeats the purpose of caching in the first place (to save JSON deserialisation).
2. We don't cache in `VersionFile.GetVersion` at all.

I'll implement 1a so you have something to review that works and implements your comments above. However, I personally think that the 2nd option (no caching) is better as it's introducing complexity and brittleness for what I imagine will be very little gain.

THanks

AArnott:
My concern about caching is that without it, You're adding *another* call to `VersionFile.GetVersion` for every commit that NB.GV has to walk in its discovery of the version height. I expect this will double the time for all NB.GV version calculations even *without* any path filters being used. That would be unacceptable as some (large) repos already experience a significant perf hit to their build from using NB.GV. We can't let this feature degrade perf for existing users.

I'll take a look at what you've done though and see how we can best mitigate it.

AArnott:
I replaced your cache work with something simpler that I think will do the trick.
At this point I think we can merge.

AArnott:
Oh whoops. I forgot about documentation. Can you write it up, @saul?
I added links to an empty `doc/pathFilters.md` file for you to fill in.

saul:
I've written documentation for this - let me know what you think.

AArnott:
Do you mind if I squash some of these commits before merging? It looks a bit noisy. :)

AArnott:
Never mind. I wouldn't want to squash *all* the commits and I guess there was a merge in there, which makes squashing certain commits a bit harder.

saul:
Any chance we could get a fresh beta release on to NuGet so we can start using this? Thanks!

AArnott:
3.1.51-beta on its way to nuget now.

cluen:
> I've written documentation for this - let me know what you think.

@saul Is it necessary to have a version.json in the monorepo root? 

I've a monorepo with these folders:

- ConsumerA/Reports/ReportA/ (containing csproj and additional sourcefiles)
- Main/ClassLibraries/Core/ (containing csproj and additional sourcefiles)
- Main/Business/Business/ (containing csproj and additional sourcefiles)


I've configured my version.json in any subdir like this.

```
{
  "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
  "version": "1.0",
  "publicReleaseRefSpec": [
    "^refs/heads/master$",
    "^refs/heads/v\\d+(?:\\.\\d+)?$"
  ],
  "pathFilters": ["./"],
  "cloudBuild": {
    "buildNumber": {
      "enabled": true
    }
  }
}
```

An update / add in any of the subfolders leads to a new version printed out by

`nbgv get-version` in that specific subfolder. 


For example: 
I've added an file to Main/ClassLibraries/Core, commited it (hash: `9dceee5c92`)

Afterwards i tried `nbgv get-version` in Main/Business/Business/
`AssemblyInformationalVersion: 1.0.5+**9dceee5c92**` 

Is this the expected behaviour or is it an configuration fault?

AArnott:
Ya, I guess we'd call this by design behavior. The path filters only keeps the version *height* value from incrementing from irrelevant changes. Nothing was ever done to the git commit that shows up for "non-public" releases to make it show the last commit that changed within that path filter.
My first thought is that this is a good design. It doesn't impact public release versions (where the git commit wouldn't show up anyway), and the whole point of it in non-public releases is to be able to determine exactly which commit it came from (rather than the last commit that changed relevant source code for it).

PR #407, opened 2019-11-17
[v3.0] Add support for ARM64 variants of Debian, Ubuntu

qmfrederik:
* Bump Microsoft.SourceLink.GitHub to the most recent version
* Add CI leg for .NET Core 3.0, on Ubuntu Bionic and Ubuntu Disco
* Add RID mapping for x64 Ubuntu 19.04, 19.10. This fixes loading of libgit2.so on these platforms.
* Add RID mapping for arm64 Debian and ARM. This enables loading of libgit2.so on these platforms.
* Update the nbgv tool to also use `GitLoaderContext`. This fixes loading of libgit2.so when invoking nbgv on the platforms listed above.


qmfrederik:
Odd, I've now seen two tests fail with `UserNotConfigured`. Not quite sure why.

qmfrederik:
- Most CI jobs are failing because LD_LIBRARY=all is set, which is causing output to stderr. They are not actually failing, though
- nbgv is failing on `ubuntu.19.04-x64`.
  * libgit2sharp should consume libgit2.so built for ubuntu.18.04-x64
  * As far I can tell, linux-x64 is being loaded (the clue is the probe for libssl1.0 instead of libssl1.1).
  * [Looking at the RID graph](https://github.com/dotnet/corefx/blob/master/src/pkg/Microsoft.NETCore.Platforms/runtime.json#L1661), it seems `ubuntu.19.04-x64` imports `ubuntu.18.10` but not `ubuntu.18.10-x64`. That might explain this.

Let's see if there's a way to make this work.

qmfrederik:
@AArnott Allright, finally got this working. In the end, it boils down to
- Updating the `RuntimeIdMap` to account for new RIDs
- Using the `GitLoaderContext` in `nbgv`, too.

CI for Ubuntu Disco is now green!

Let me know if you think this can be merged, looking forward to (finally!) using nbvg on my Raspberry Pi 4 :smile: .

azure-pipelines/xplattest-pipeline.yml
```diff
@@ -20,7 +20,7 @@ steps:
 
     NBGV_NuGetPackageVersion=$([[ $PkgFileName =~ Nerdbank.GitVersioning\.(.*)\.nupkg ]] && echo "${BASH_REMATCH[1]}")
 
-    echo "<configuration><packageSources><clear /><add key='local' value='deployables' /></packageSources></configuration>" > nuget.config &&
+    echo "<configuration><packageSources><add key='local' value='deployables' /></packageSources></configuration>" > nuget.config &&
```
AArnott: Why this change? We cleared the feeds so we don't restore anything we didn't just build.

src/NerdBank.GitVersioning.NativeBinaries/NerdBank.GitVersioning.NativeBinaries.csproj
```diff
@@ -0,0 +1,13 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- Download version 2.0.280-armpreview01 of LibGit2Sharp.NativeBinaries. This builds from the same libgit2 commit (git2-572e4d8) as 
+         0.27.0-preview-0007, but includes arm64 binaries.
+         This explicit dependency should be removed when upgrading to a newer version of LibGit2Sharp. -->
+    <PackageReference Include="LibGit2Sharp.NativeBinaries" Version="2.0.280-armpreview01" />
```
AArnott: Is this reference the entire reason for this project? 
If so, can we simply move it into an existing project (and set `PrivateAssets="all"` on it) or use [the `PackageDownload` item](https://github.com/NuGet/NuGet.Client/blob/0ab64e658f3c266e8e805da92dd83e0fa26280ed/build/bootstrap.proj#L11)?

src/Nerdbank.GitVersioning.Tasks/GitLoaderContext.cs
```diff
@@ -15,6 +15,11 @@ internal class GitLoaderContext : AssemblyLoadContext
     {
         public static readonly GitLoaderContext Instance = new GitLoaderContext();
 
+        // When invoked as a MSBuild task, the native libraries will be at
+        // ../runtimes. When invoked from the nbgv CLI, the libraries
+        // will be at ./runtimes
```
AArnott: Can we beef up this comment to explain where this property gets set to `./runtimes`? I finally found it elsewhere in this PR, but just looking at the comment it is not at all clear.

azure-pipelines/xplattest-pipeline.yml
qmfrederik: Without this, [restore fails when using the .NET Core 3.0 SDK](https://andrewarnott.visualstudio.com/OSS/_build/results?buildId=1959):

```
Running 'dotnet restore' on lib/lib.csproj...
/__w/1/s/lib/lib.csproj : error NU1101: Unable to find package NETStandard.Library. No packages exist with this id in source(s): local
  Restore failed in 120.28 ms for /__w/1/s/lib/lib.csproj.
```

I guess it's trying to resolve NuGet packages from the local NuGet feed, but it can't, since that feed has been cleared.

master also has a similar change, via #381

src/NerdBank.GitVersioning.NativeBinaries/NerdBank.GitVersioning.NativeBinaries.csproj
qmfrederik: Oh, I didn't know about the `PackageDownload` item. Looks much cleaner now. Thanks!

src/Nerdbank.GitVersioning.Tasks/GitLoaderContext.cs
qmfrederik: Sure, I updated the comments.

src/NerdBank.GitVersioning.NativeBinaries/NerdBank.GitVersioning.NativeBinaries.csproj
qmfrederik: On the downside, `PackageDownload` works start with .NET SDK 2.2 which somehow broke the layout of the `nbgv.cli` folder used by the npm package. I replaced the copy mechanism with the output of `dotnet publish`, that seems to do the trick.

qmfrederik:
@AArnott. I think I adressed your comments. Let me know if there's anything else you need!

PR #431, opened 2020-01-22
Add a ReleaseOptions.VersionFieldCount option to control version in branch name

remcoros:
This allows specifying the number of version components to use in the branch name.

Example setup:

```json
{
  "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
  "version": "1.0.0-preview.{height}",
  "nuGetPackageVersion": {
    "semVer": 2.0
  },
  "publicReleaseRefSpec": [
    "^refs/heads/master$",
    "^refs/heads/release/v\\d+(?:\\.\\d+)?$"
  ],
  "release": {
    "branchName": "release/v{version}",
    "firstUnstableTag": "preview"
  }
}
```

- We don't want to use {height} in public release number (use version with 3 components)
- Minor releases are managed from their own branch (release/v1.0 release/v1.1 release/v2.0 etc)
- version in a release branch is: "1.0.0" and manually updated to "1.0.1-preview.{height}" or "1.0.1" when releasing (preview)patches.

Currently when using this setup, we cannot use "nbgv prepare-release", as this creates a branch named "release/v1.0.0"

The "VersionFieldCount" option allows to override how many fields of the version should be used when constructing the branch name.

This would be my new settings file:

```json
{
  "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
  "version": "1.0.0-preview.{height}",
  "nuGetPackageVersion": {
    "semVer": 2.0
  },
  "publicReleaseRefSpec": [
    "^refs/heads/master$",
    "^refs/heads/release/v\\d+(?:\\.\\d+)?$"
  ],
  "release": {
    "branchName": "release/v{version}",
    "firstUnstableTag": "preview",
    "versionFieldCount": 2
  }
}
```

Notes:

- I'm not sure I like the "VersionFieldCount" option name, but couldn't come up with a better one.

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/ReleaseManagerTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/ReleaseManagerTests.cs
@@ -86,6 +86,27 @@ public class ReleaseManagerTests : RepoTestBase
     }
 
     [Fact]
+    public void PrepareRelease_InvalidVersionFieldCountSetting()
+    {
+        this.InitializeSourceControl();
+
+        // create version.json
+        var versionOptions = new VersionOptions()
+        {
+            Version = SemanticVersion.Parse("1.2-pre"),
+            Release = new ReleaseOptions()
+            {
+                VersionFieldCount = 3
+            }
+        };
+        this.WriteVersionFile(versionOptions);
+
+        // running PrepareRelease should result in an error
+        // because the branchName does not have a placeholder for the version
+        this.AssertError(() => new ReleaseManager().PrepareRelease(this.RepoPath), ReleasePreparationError.InvalidVersionFieldCountSetting);
+    }
+
+    [Fact]
     public void PrepareRelease_ReleaseBranchAlreadyExists()
     {
         this.InitializeSourceControl();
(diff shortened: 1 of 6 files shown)


AArnott:
Thanks for contributing, but opening an issue to propose your idea first can save you some potentially wasted effort. NB.GV does not let branch names or tags influence the versioning of the project (except to influence the `IsPublicRelease` boolean value based on `publicReleaseRefSpec`. This design decision is important for build reproducibility since a given commit might (and very likely will) appear on many branches and thus could build as v1.0 and v1.5 depending on what branch is 'checked out' at the time. It's an intentional design decision that the version has to be burned into the commit itself by touching version.json's `version` property.

remcoros:
@AArnott Thanks for the quick response!

I wanted to get familiar with the inner workings anyway, so this what not a wasted effort for me, but very fun.

I think tough, you understand my question wrong. But I will rephrase and explain in a separate issue, so let's continue there.

PR #436, opened 2020-02-13
Changing the hardcoded `g` prefix 

vchirikov:
Now we can change the hardcoded `g` prefix.

### DIFF
--- a/src/NerdBank.GitVersioning/VersionOracle.cs
+++ b/src/NerdBank.GitVersioning/VersionOracle.cs
@@ -409,7 +409,7 @@
         /// See <see href="https://github.com/AArnott/Nerdbank.GitVersioning/issues/260#issuecomment-445511898">this discussion</see>.
         /// </remarks>
         private string NuGetSemVer1BuildMetadata =>
-            this.PublicRelease ? string.Empty : $"-g{this.GitCommitIdShort}";
+            this.PublicRelease ? string.Empty : $"-{this.VersionOptions?.GitCommitIdPrefix ?? "g"}{this.GitCommitIdShort}";
 
         /// <summary>
         /// Gets the build metadata, compliant to SemVer 1.0.
@@ -418,7 +418,7 @@
             this.PublicRelease ? string.Empty : $"-{this.GitCommitIdShort}";
 
         /// <summary>
-        /// Gets a SemVer 1.0 compliant string that represents this version, including the -gCOMMITID suffix
+        /// Gets a SemVer 1.0 compliant string that represents this version, including the -{GitCommitIdPrefix}COMMITID suffix
         /// when <see cref="PublicRelease"/> is <c>false</c>.
         /// </summary>
         private string NuGetSemVer1 =>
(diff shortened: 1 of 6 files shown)


AArnott:
Thanks for your contribution. But why? 

(Tip: It's a good idea to file an issue proposing your change so we can discuss it before you go to the work of implementing it).

vchirikov:
I don't like the hardcoded `g` prefix for git commit, it can confuse newbies.
For example I want to use semver2.0 and `git.` prefix. (I don't need to support old nuget clients).
p.s. Also could you explain [this change](https://github.com/AArnott/Nerdbank.GitVersioning/blob/master/src/NerdBank.GitVersioning/VersionOracle.cs#L447) `.` to `-` in prerelease version?
p.p.s we can discuss here :) btw thanks for your work.

AArnott:
> For example I want to use semver2.0 and git. prefix. (I don't need to support old nuget clients).

Fair enough.

> p.s. Also could you explain this change . to - in prerelease version?

Yes: the git suffix is the last identifier in the unstable tag. Per semver2 rules:

>a hyphen and a series of dot separated identifiers

Therefore if our git commit id happens to be the first identifier, it must be preceded by a hyphen. Othrwise, it should be preceded by a period.

src/NerdBank.GitVersioning/VersionOracle.cs
```diff
@@ -438,13 +438,13 @@ public IEnumerable<string> BuildMetadataWithCommitId
         private string PrereleaseVersionSemVer1 => SemanticVersionExtensions.MakePrereleaseSemVer1Compliant(this.PrereleaseVersion, this.SemVer1NumericIdentifierPadding);
 
         /// <summary>
-        /// Gets the -gc0ffee or .gc0ffee suffix for the version.
+        /// Gets the -{<seealso cref="VersionOptions.GitCommitIdPrefix"/>}c0ffee or .<seealso cref="VersionOptions.GitCommitIdPrefix"/>c0ffee suffix for the version.
```
AArnott: Same here. Let's just keep `g` as the prefix here.
You could add a parenthetical explanation like "(where the g in the prefix might be changed if `<see cref="VersionOptions.GitCommitIdPrefix"/>` is set."

src/NerdBank.GitVersioning/version.schema.json
```diff
@@ -79,6 +79,12 @@
           "default": 10,
           "maximum": 40
         },
+        "gitCommitIdPrefix": {
+          "type": "string",
+          "description": "The git commit prefix (e.g. 'g') in prerelease versions",
```
AArnott: Change description to:

> The git commit prefix (e.g. 'g') in non-public release versions.

The reason is the switch for whether we include the git commit ID in the version isn't whether it's prerelease or not. It's whether it's built as a public release or not.

AArnott:
Closing due to no response.

vchirikov:
> Closing due to no response.

Sorry for late response, I've fixed code according to your notices.

AArnott:
Thanks. I would reopen but GitHub won't let me. Can you open a new pull request?

vchirikov:
okay

Issue #414, opened 2019-12-04
Builds broken in .NET SDK Alpine Image 


fgreinacher:
I am seeing assembly load errors (see below) when referencing the NBGV NuGet package in a project and building it within [the official Alpine image of the .NET Core 3.1 SDK](https://github.com/dotnet/dotnet-docker/blob/d4a9e799d047f3e86cd2730f48b689c371d38480/3.1/sdk/alpine3.10/amd64/Dockerfile).

I prepared a [tiny repro repo](https://github.com/fgreinacher/nbgv-alpine-issue) (🤣).

Any help (including workarounds) would be highly appreciated 👍 

```
/root/.nuget/packages/nerdbank.gitversioning/3.0.28/build/Nerdbank.GitVersioning.targets(71,5): error MSB4018:    at System.Runtime.Loader.AssemblyLoadContext.ResolveUnmanagedDll(String unmanagedDllName, IntPtr gchManagedAssemblyLoadContext) [/root/project/nbgv-alpine-issue.csproj]
/root/.nuget/packages/nerdbank.gitversioning/3.0.28/build/Nerdbank.GitVersioning.targets(71,5): error MSB4018:    at LibGit2Sharp.Core.NativeMethods.git_libgit2_init() [/root/project/nbgv-alpine-issue.csproj]
/root/.nuget/packages/nerdbank.gitversioning/3.0.28/build/Nerdbank.GitVersioning.targets(71,5): error MSB4018:    at LibGit2Sharp.Core.NativeMethods.InitializeNativeLibrary() [/root/project/nbgv-alpine-issue.csproj]
/root/.nuget/packages/nerdbank.gitversioning/3.0.28/build/Nerdbank.GitVersioning.targets(71,5): error MSB4018:    at LibGit2Sharp.Core.NativeMethods..cctor() [/root/project/nbgv-alpine-issue.csproj]
/root/.nuget/packages/nerdbank.gitversioning/3.0.28/build/Nerdbank.GitVersioning.targets(71,5): error MSB4018:    --- End of inner exception stack trace --- [/root/project/nbgv-alpine-issue.csproj]
/root/.nuget/packages/nerdbank.gitversioning/3.0.28/build/Nerdbank.GitVersioning.targets(71,5): error MSB4018:    at LibGit2Sharp.Core.NativeMethods.git_libgit2_opts(Int32 option, UInt32 level, String path) [/root/project/nbgv-alpine-issue.csproj]
[...]

PR #440, opened 2020-03-03
Update libgit2sharp to support alpine 3.9 rid

woodne:


woodne:
This should address the issue in #414, at least for alpine 3.9.

PR #442, opened 2020-03-05
Hide ExcludeFromCodeCoverageAttribute usage on nanoFramework

### LINKED ISSUE #441: New CodeAnalysis attribute breaking nanoFramework builds
Hi,

Recently this line was added:
https://github.com/AArnott/Nerdbank.GitVersioning/blob/35b087bc50864c8da8b9abfd2124f59ee31a74dd/src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs#L545

Bring the `System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage` attribute.
Because we don't have that attribute in nanoFramework this is breaking our builds.

If you recall we had a previous contact in #346 because of a similar problem.

Any chance that can be wrapped with compiler defs?

AArnott:
Fixes #441


AArnott:
@josesimoes Can you test this?

josesimoes:
It's failing the test but from what I can see in the code you are now adding this only for those platforms, which is OK for us.

josesimoes:
Just pulled the latest beta NuGet and the build went fine. 💯 👍

PR #446, opened 2020-03-09
Code sign in CI builds

AArnott:
CC: @onovotny

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -28,6 +28,8 @@ resources:
 jobs:
 - job: Windows
   pool: Hosted Windows 2019 with VS2019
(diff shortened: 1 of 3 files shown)


onovotny:
One suggestion is to split the build and sign into separate stages (like the sample repo) to minimize any possible exposure of the secrets during the build. Since the secrets in the variable group aren't in the build phase, things like binlogs won't accidently contain them.

AArnott:
It's a fair point, @onovotny. But I'm using a 'secret' variable anyway, so it's not available to any tool unless I explicitly pass it in from the YAML file. You can see in the binlog file that only the username is available.

image

PR #470, opened 2020-04-27
Set BuildingRef as output property from GetBuildVersion

aok-foss:
It's not a good idea to take a dependency on a branch name. At the same time, branches are too useful to ignore. The `PublicRelease` functionality depends on them.
The argument can go one step further: during development (in non-public releases), branch names in version numbers are useful for traceability: e.g. `1.0.0-g1234567+PR1512` or `1.0.0-g1234567+BalpineLinux`

It would be easiest to extend NBGV-produced version numbers in the way above if NBGV also provided access to build refs.

### DIFF
--- a/src/Nerdbank.GitVersioning.Tasks/GetBuildVersion.cs
+++ b/src/Nerdbank.GitVersioning.Tasks/GetBuildVersion.cs
@@ -21,11 +21,6 @@
         }
 
         /// <summary>
-        /// Gets or sets the ref (branch or tag) being built.
-        /// </summary>
-        public string BuildingRef { get; set; }
-
-        /// <summary>
         /// Gets or sets identifiers to append as build metadata.
         /// </summary>
         public string[] BuildMetadata { get; set; }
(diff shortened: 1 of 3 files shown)


src/Nerdbank.GitVersioning.Tasks/GetBuildVersion.cs
```diff
@@ -20,11 +20,6 @@ public GetBuildVersion()
         {
         }
 
-        /// <summary>
-        /// Gets or sets the ref (branch or tag) being built.
-        /// </summary>
-        public string BuildingRef { get; set; }
```
AArnott, line 26: Removing this seems irrelevant, and would be a breaking change and drop support for an important use case.

aok-foss: Thought I was removing some dead code. Will restore.

src/Nerdbank.GitVersioning.Tasks/GetBuildVersion.cs
```diff
@@ -78,6 +78,13 @@ public GetBuildVersion()
         [Output]
         public bool PublicRelease { get; private set; }
 
+        /// <summary>
+        /// Gets the ref (branch or tag) being built.
+        /// Only specified when this is not a public release.
+        /// </summary>
+        [Output]
+        public string PrivateReleaseGitRef { get; set; }
```
AArnott, line 86: I don't see any benefit to only setting this when it's not a public release. We also have a `PublicRelease` output parameter so folks can condition their own code. Do you object to just making this `BuildingRef` and setting it unconditionally?

aok-foss: No objections. Updated.

PR #476, opened 2020-05-12
Add Arch Linux runtime ID

JamieMagee:
The Arch Linux runtime id was added to the .NET runtime in [this commit](https://github.com/dotnet/runtime/commit/00d8f8e378be27edeb37a4f5655ccd831f1cf891), which suggests that it should be compatible with the `linux-x64` native dependencies.

I would appreciate some input on the best way to test this out locally


AArnott:
Thanks for contributing!

Testing locally would involve a docker container. 
And similarly, what I would want with this PR is the addition of a new test job in our Azure Pipeline to help ensure that it works and keeps working. Can you add that test job? You can follow the pattern here: 

https://github.com/dotnet/Nerdbank.GitVersioning/blob/d46ca3affd480dbda2914537d29b1a5ba236c0db/azure-pipelines.yml#L188-L195

JamieMagee:
@AArnott So I can test this locally, but I'm struggling with the CI.

I'm trying to run `pacman -Sy --noconfirm git dotnet-sdk` and this passes locally, but when running in Azure DevOps it gives the error

> you cannot perform this operation unless you are root.

But the ArchLinux docker image doesn't ship with `sudo` by default. I think it might be the same issue as microsoft/azure-pipelines-agent#2043.

Do you have any suggestions? Or would you accept a PR if I tried this on GitHub Actions?

AArnott:
I have similarly complained when docker images lacked a `sudo` command and Azure Pipelines didn't give you root access. 
I'll take a look.
Worst case scenario: we add a layer to that Arch Linux image that adds sudo and consume that from this pipeline.

AArnott:
With my added commit, we're getting farther. But it fails likely due to missing native dependencies. Did you say it's working in your local testing? What prereqs are you adding to the image when it works for you?

JamieMagee:
I was missing the `openssl-1.0` package. Thanks for the help! It works now :smile:

JamieMagee:
@AArnott Do you have time today to take another look at this?

JamieMagee:
@AArnott thanks for merging. Is it possible to release a stable version to NuGet with this fix?

AArnott:
I can, but I tend to leave prereleases open for longer to collect more new features.
Is there some reason you need a stable version as opposed to a prerelease on nuget.org? It shouldn't impact any of your downstream users since it's a `PrivateAssets="all"` package reference that doesn't appear to your own users.

JamieMagee:
I think we'd prefer to depend on a stable version, rather than a pre-release version. But, I'll open a PR and get some feedback on if a beta version is acceptable.

AArnott:
NB.GV is generally "stable" nowadays. I don't think we'll do anything to really destabilize without a major version bump at this point.

PR #486, opened 2020-05-29
Add JSON output option to 'prepare-release' command

### LINKED ISSUE #481: Make output of "prepare-release" command machine-readable
I'm running the `prepare-release` command as part of a script that performs other actions required for release.

The script needs the name of the branch that `prepare-release` created. 
Currently this means either 

 - parsing the log output or 
- duplicating the logic to determine the name of the branch from the settings in `version.json` in the script.

I this having an option to output the information in a machine-readable way would be preferable.

I propose to add a `--format` option to the `prepare-release` command that emits the relevant information as json (analogous to the `--format` option of  `get-version`).

Would you accept a PR adding this feature?

The output could look something like this (branching off `0.2` from `master`):

```json
{
    "releaseBranch" : {
        "name" : "release/v0.2",
        "version" : "0.2"
    },
    "devBranch" : {
       "name" : "master", 
       "version" : "0.3-pre"
    }
}
```

ap0llo:
This PR adds a `--format` parameter to `prepare-release` that allows setting the output format to `json`.

I made some adjustments to the output format proposed in #481. 
The output will look like this:

```json
{
  "CurrentBranch": {
    "Name": "master",
    "Commit": "5a7487098ac1be1ceb4dbf72d862539cf0b0c27a",
    "Version": "0.3-pre"
  },
  "NewBranch": {
    "Name": "release/v0.2",
    "Commit": "b2f164675ffe891b66b601c00efc4343581fc8a5",
    "Version": "0.2"
  }
}
```
- The output includes the id of the latest commit on the respective branches. This way, the exact version on the branch can easily be queried using `get-version`.
- The property names use pascal-case instead of camel-case so the output is consistent with the JSON output of  `get-version`
- The properties `devBranch` and `releaseBranch` are named`CurrentBranch`(the branch the command was started on) and `NewBranch` (the new branch that was created) instead. 
  That way, the scenario that the command is started on the release branch can be handled as well. In that case, the `NewBranch` property will be `null` because no new branch was created.

Fixes #481

### DIFF
--- a/src/NerdBank.GitVersioning/ReleaseManager.cs
+++ b/src/NerdBank.GitVersioning/ReleaseManager.cs
@@ -115,11 +115,7 @@
             [JsonConstructor]
             public ReleaseInfo(ReleaseBranchInfo currentBranch, ReleaseBranchInfo newBranch)
             {
-                if(currentBranch == null)
-                {
-                    throw new ArgumentNullException(nameof(currentBranch), $"{nameof(this.CurrentBranch)} can't be empty");
-                }
-                
+                Requires.NotNull(currentBranch, nameof(currentBranch));                
                 // skip null check for newBranch, it is allowed to be null.
 
                 this.CurrentBranch = currentBranch;                
@@ -155,20 +151,9 @@
             /// <param name="version">The version configured in the branch's <c>version.json</c>.</param>
             public ReleaseBranchInfo(string name, string commit, SemanticVersion version)
             {
-                if (string.IsNullOrWhiteSpace(name))
-                {
-                    throw new ArgumentNullException(nameof(name), $"{nameof(this.Name)} can't be empty");
-                }
-
-                if (string.IsNullOrWhiteSpace(commit))
-                {
-                    throw new ArgumentNullException(nameof(commit), $"{nameof(this.Commit)} can't be empty");
-                }
-
-                if(version == null)
-                {
-                    throw new ArgumentNullException(nameof(version), $"{nameof(this.Version)} can't be empty");
-                }
+                Requires.NotNullOrWhiteSpace(name, nameof(name));
+                Requires.NotNullOrWhiteSpace(commit, nameof(commit));
+                Requires.NotNull(version, nameof(version));
 
                 this.Name = name;
                 this.Commit = commit;
--- a/src/NerdBank.GitVersioning/ReleaseManager.cs
+++ b/src/NerdBank.GitVersioning/ReleaseManager.cs
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.IO;
     using LibGit2Sharp;
+    using Newtonsoft.Json;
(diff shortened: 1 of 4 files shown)


AArnott:
> The property names use camel-case instead of pascal-case so the output is consistent with the JSON output of get-version

Your sample actually shows PascalCase. So I'm confused as to which one is accurate.

src/NerdBank.GitVersioning/ReleaseManager.cs
```diff
@@ -80,11 +81,122 @@ public class ReleasePreparationException : Exception
...
+                {
+                    throw new ArgumentNullException(nameof(name), $"{nameof(this.Name)} can't be empty");
+                }
+
+                if (string.IsNullOrWhiteSpace(commit))
+                {
+                    throw new ArgumentNullException(nameof(commit), $"{nameof(this.Commit)} can't be empty");
+                }
+
+                if(version == null)
```
AArnott, line 168: nit: your whitespace formatting is inconsistent. Please execute Format Document on each file you change to normalize it.

ap0llo:
> 
> 
> > The property names use camel-case instead of pascal-case so the output is consistent with the JSON output of get-version
> 
> Your sample actually shows PascalCase. So I'm confused as to which one is accurate.

Sorry about that, I wrote that down in the wrong order.
I fixed the PR description

src/NerdBank.GitVersioning/ReleaseManager.cs
```diff
@@ -80,11 +81,122 @@ public class ReleasePreparationException : Exception
...
+            /// Initializes a new instance of <see cref="ReleaseBranchInfo"/>.
+            /// </summary>
+            /// <param name="name">The name of the branch.</param>
+            /// <param name="commit">The id of the branch's tip.</param>
+            /// <param name="version">The version configured in the branch's <c>version.json</c>.</param>
+            public ReleaseBranchInfo(string name, string commit, SemanticVersion version)
+            {
+                if (string.IsNullOrWhiteSpace(name))
+                {
+                    throw new ArgumentNullException(nameof(name), $"{nameof(this.Name)} can't be empty");
```
AArnott, line 160: nit: don't throw `ArgumentNullException` for non-null conditions like an empty string.
tip: Replace all your "if-throw" checks with `Requires.NotNull` or `Requires.NotNullOrEmpty` method calls, which handle the messages and exception types to be thrown.

src/NerdBank.GitVersioning/ReleaseManager.cs
ap0llo: Sure, whitespace issues should be resolved now.

src/NerdBank.GitVersioning/ReleaseManager.cs
ap0llo: I updated the argument checks to use `Requires`.

PR #501, opened 2020-08-31
Add name of cloudbuild environment variable for TeamCity

pergardebrink:
The TeamCity documentation didn't state what the name of the environment variable was supposed to be to get it working. I got the name from the following line: 

https://github.com/dotnet/Nerdbank.GitVersioning/blob/44e8bfc8dc6ae700509563b0bf1d9766b1188ed1/src/NerdBank.GitVersioning/CloudBuildServices/TeamCity.cs#L16

### DIFF
--- a/doc/cloudbuild.md
+++ b/doc/cloudbuild.md
@@ -149,7 +149,7 @@ Check out [nerdbank-gitversioning on the GitHub Actions marketplace](https://git
 
 ### TeamCity
 TeamCity does not expose the build branch by default as an environment variable. This can be exposed by
-adding an environment variable with the value of `%teamcity.build.vcs.branch.<vcsid>%` where `<vcsid>` is
+adding an environment variable called BUILD_GIT_BRANCH with the value of `%teamcity.build.vcs.branch.<vcsid>%` where `<vcsid>` is
 the root id described on the TeamCity VCS roots page. Details on this variable can be found on the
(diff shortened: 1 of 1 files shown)


pergardebrink:
Been banging my head against this for a while now that this and my other PR doesn't build because the BuildIntegrationTests are failing complaining about System.Runtime.CompilerServices.Unsafe. 

I get the same error locally, but can get it to run tests correctly if I the following versions are added in Nerdbank.GitVersioning.Tests.csproj: 
``` 
<PackageReference Include="System.Numerics.Vectors" Version="4.4.0" />
<PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.3" />
```

Is it a transitive dependency that for some reason is blocked? Cannot see how that would happen..

pergardebrink:
Someone else is having a similar problem with msbuild 16.7 (same as I have): https://github.com/xamarin/xamarin-android/pull/4995/files

Maybe we also need explicit binding redirects here from the test project?

AArnott:
> Been banging my head against this for a while now that this and my other PR doesn't build because the BuildIntegrationTests are failing complaining about System.Runtime.CompilerServices.Unsafe.

I just requeued a CI and it's failing too. So it's not your fault. I'll see what I can do to get it working again.

AArnott:
The failure is due to what I believe is a bug in the MSBuild Locator: https://github.com/microsoft/MSBuildLocator/issues/95

But I can workaround it.

PR #507, opened 2020-09-04
Perf: Cache JSON parse results while computing height

AArnott:
Previous to this, we were re-parsing the same JSON content for every commit we encountered it in. Now we cache it based on the blob ID so (except in certain inherit cases) we never parse a given JSON content more than once while computing height.

As part of this caching I had to make sure that a cached `VersionOptions` instance would not be changed later, which is exactly what happens when `inherit: true`, so most of the diff is me making this type and all other types in the object tree freezable so it'll throw instead of allow corruption.

### DIFF
--- a/src/NerdBank.GitVersioning/GitExtensions.cs
+++ b/src/NerdBank.GitVersioning/GitExtensions.cs
@@ -984,6 +984,7 @@
         private class GitWalkTracker
         {
             private readonly Dictionary<ObjectId, VersionOptions> commitVersionCache = new Dictionary<ObjectId, VersionOptions>();
+            private readonly Dictionary<ObjectId, VersionOptions> blobVersionCache = new Dictionary<ObjectId, VersionOptions>();
             private readonly Dictionary<ObjectId, int> heights = new Dictionary<ObjectId, int>();
 
             internal GitWalkTracker(string repoRelativeDirectory)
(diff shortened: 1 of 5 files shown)


AArnott:
@djluck @filipnavara @qmfrederik I'd welcome your measurements on perf impact of this change on your repos.

djluck:
@AArnott this has made a big improvement to perf. Previously, calls to `VersionFile.GetVersion` were accounting for about ~30-50% of our runtime and now the balance has drastically shifted:
image

Now we can see the majority of the time is spent:
- Performing diffs (52.3%)
- Fetching configuration values from git (28.7%)

The configuration fetch cost seems to stem from reading the `core.ignorecase` configuration property per commit- not sure if there's an easy fix for this (I guess you want to respect any changes made to the repository git configuration) and to my knowledge, this will only impact building on windows (windows isn't optimized for fetching files by name repetitively like git does).

As for git diffing, there might be a quick win we can apply here. The perf trace shows that 32.2% of time is spent in `git_diff_find_similar`- by passing the following options to the diff method:

```
var compareOptions = new CompareOptions() { ContextLines = 0, Similarity = SimilarityOptions.None };
```
We can reduce the amount of time we spend diffing as we're not checking for moved or renamed files. To my mind, we don't need to do this as we only care if a file was added or removed in a path to bump the version height. A rough benchmark showed it reduced average time to calculate the version height by 25%.

AArnott:
Thanks for the positive feedback and ideas for further improvement, @djluck. Your proposal to improve diff time sounds great. Can you send a PR?

djluck:
Sure thing, it should be an easy change. I'm out over the weekend but I'll sort something out for next week.

PR #513, opened 2020-09-24
Reducing CPU consumption when calculating git versions by not checkin…

djluck:
…g files for similarities


djluck:
@AArnott as requested, PR to add discussed changes around diff generations.

AArnott (approved):
Thanks

PR #516, opened 2020-10-05
Github Actions: Use Environment File

talenfisher:
Recently, Github Actions [deprecated the set-env command in favor of using environment files](https://github.com/actions/toolkit/security/advisories/GHSA-mfwh-5m23-j46w).  Since then, I've been getting a lot of warnings about set-env that stem from its use here.   

This PR updates GithubActions.SetCloudBuildVariable to write to the environment file, rather than use the set-env command.

### DIFF
--- a/src/NerdBank.GitVersioning/CloudBuildServices/GitHubActions.cs
+++ b/src/NerdBank.GitVersioning/CloudBuildServices/GitHubActions.cs
@@ -28,7 +28,7 @@
 
         public IReadOnlyDictionary<string, string> SetCloudBuildVariable(string name, string value, TextWriter stdout, TextWriter stderr)
         {
-            File.AppendAllText(EnvironmentFile, $"{name}={value}{Environment.NewLine}");
+            File.AppendAllText(EnvironmentFile, $"{Environment.NewLine}{name}={value}{Environment.NewLine}");
             return GetDictionaryFor(name, value);
         }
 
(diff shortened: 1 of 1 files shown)


src/NerdBank.GitVersioning/CloudBuildServices/GitHubActions.cs
```diff
@@ -19,14 +19,16 @@ internal class GitHubActions : ICloudBuild
...
+
         public IReadOnlyDictionary<string, string> SetCloudBuildNumber(string buildNumber, TextWriter stdout, TextWriter stderr)
         {
             return new Dictionary<string, string>();
         }
 
         public IReadOnlyDictionary<string, string> SetCloudBuildVariable(string name, string value, TextWriter stdout, TextWriter stderr)
         {
-            (stdout ?? Console.Out).WriteLine($"##[set-env name={name};]{value}");
+            File.AppendAllText(EnvironmentFile, $"{name}={value}{Environment.NewLine}");
```
AArnott, line 31: Would it be prudent to add a newline at the start of this string as well in case the last person to append to the file didn't add a newline character themselves?

AArnott (changes requested):
Thanks for contributing! This looks like a good fix. One quick suggestion. What do you think?

src/NerdBank.GitVersioning/CloudBuildServices/GitHubActions.cs
talenfisher: I think thats a good idea, done in 795f5df

AArnott (approved):
Thanks

Issue #505, opened 2020-09-04
Consider using a managed Git backend


qmfrederik:
### TL;DR

I've done some experiments with a managed Git backend (instead of libgit2).

In all scenarios, performance is better with the managed backend. In the best case, throughput increases more than tenfold.
The prototype source code is available here: https://github.com/qmfrederik/quamotion.gitversioning.

There are other managed Git backends ([GitNet](https://github.com/choffmeister/gitnet), [GitSharp](https://github.com/henon/GitSharp), [NGit](https://github.com/mono/ngit)) available.

|                             Backend |             Repository |      Mean |     Error |    StdDev |
|----------------------------------- |--------------------- |----------:|----------:|----------:|
| managed |  Cuemon |  6.585 ms | 0.0977 ms | 0.0914 ms |
|        libgit2|  Cuemon | 12.946 ms | 0.1425 ms | 0.1333 ms |
| managed | SuperSocket |  7.583 ms | 0.1067 ms | 0.0998 ms |
|        libgit2| SuperSocket | 61.295 ms | 0.8168 ms | 0.7640 ms |
| managed | local | 1.938 ms | 0.0161 ms | 0.0151 ms |
|        libgit2| local | 24.805 ms | 0.3896 ms | 0.3253 ms |
| managed |   xunit| 57.891 ms | 0.6076 ms | 0.5386 ms |
|        libgit2|   xunit | 62.054 ms | 0.9426 ms | 0.7871 ms |

### Why

NerdBank.GitVersioning uses LibGit2Sharp as its back-end. It comes with a couple of drawbacks:

- Performance - libgit2 is a general-purpose library and perhaps not geared towards read-only scenarios like nbgv, P/Invoke overhead,... .
- Maintainability - it looks like development on LibGit2Sharp has stalled a bit (last commit is from April this year)
- Portability - there's [a very long list](https://github.com/dotnet/Nerdbank.GitVersioning/issues?q=is%3Aissue+is%3Aclosed+label%3Alibgit2) of issues related to nbgv not on various Linux distributions

### What

My goal was to implement a minimal viable Git backend which you can use to calculate the Git height. That's all.

This includes
- Read commits, trees and blobs from a local Git repository
- Support for 'packed' Git repositories (i.e. what you get when you call `git gc` or after a fresh Git clone) and deltafied objects
- In-memory caching

I've also applied some of the suggestions related to performance made by @filipnavara  and @djluck, such as
[...]

PR #520, opened 2020-10-18
Enable tests on netcoreapp3.1

qmfrederik:
I'm prototyping work related to #505. The managed Git implementation, at the moment, uses `Span<T>` and friends, and targets `netcoreapp3.1`.

I retargeted the project and tests to run on `netcoreapp3.1` and came across a couple of test failures, which this PR attempts to resolve:

- The `AssemblyVersionInfo` task throws no error when targeting a language which is not supported
- `RuntimePath` appears to be `./runtimes` instead of `../runtimes`
- SevenZipNET does not target `netcoreapp`, and is only used in a test which is skipped. I have methods which rely on SevenZipNET throw a `PlatformNotSupportException`.

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -940,6 +940,7 @@ public class BuildIntegrationTests : RepoTestBase, IClassFixture<MSBuildFixture>
         Assert.Empty(result.LoggedEvents.OfType<BuildWarningEventArgs>());
     }
 
+#if !NETCORE
     /// <summary>
     /// Create a native resource .dll and verify that its version
     ///  information is set correctly.
@@ -962,6 +963,7 @@ public class BuildIntegrationTests : RepoTestBase, IClassFixture<MSBuildFixture>
         Assert.Equal("NerdBank", fileInfo.CompanyName);
         Assert.Equal($"Copyright (c) {DateTime.Now.Year}. All rights reserved.", fileInfo.LegalCopyright);
     }
+#endif
 
     private static Version GetExpectedAssemblyVersion(VersionOptions versionOptions, Version version)
     {
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -940,7 +940,7 @@ public class BuildIntegrationTests : RepoTestBase, IClassFixture<MSBuildFixture>
         Assert.Empty(result.LoggedEvents.OfType<BuildWarningEventArgs>());
     }
 
-#if !NETCORE
+#if !NETCOREAPP
     /// <summary>
     /// Create a native resource .dll and verify that its version
     ///  information is set correctly.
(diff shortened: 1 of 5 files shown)


qmfrederik:
@AArnott - `NativeVersionInfo_CreateNativeResourceDll` still fails. Do you know if MSBuild Core supports `vcxproj` projects?

AArnott:
`NativeVersionInfo_CreateNativeResourceDll` fails locally for me too. MSBuild certainly does support vcxproj, but I don't remember why it fails locally. But [it passes in CI](https://dev.azure.com/andrewarnott/OSS/_build/results?buildId=3607&view=ms.vss-test-web.build-test-results-tab&runId=1022670&resultId=100016&paneView=debug).

AArnott:
That said, interestingly that vcxproj test passes on net461 in CI but [fails on netcoreapp3.1](https://dev.azure.com/andrewarnott/OSS/_build/results?buildId=3637&view=logs&j=0bc77094-9fcd-5c38-f6e4-27d2ae131589&t=ea548fe5-251b-5eea-f78a-7ecee48fb1fc). If we can't figure out how to fix it, we can disable it using something like this on the test method:

```cs
#if NETCOREAPP
[Trait("TestCategory", "FailsOnAzurePipelines")]
#endif
```

qmfrederik:
93d9d305d0c8a74d1cdd5bade51469f0aea345ba might do the trick by making the test conditional - the test won't work if the v142 toolset is not installed or the Windows 10 SDK could not be resolved.

I don't fully understand why the test would succeed on NetFx and fail on NetCore, though, perhaps somebody with a deeper understanding of the VC toolset can explain (or we just skip the test on Azure, like you suggest) :).

AArnott:
I suspect it's because under .NET Core, the msbuild installation comes from the .NET Core SDK, which doesn't have C++ extensions installed. But when run under .NET Framework, the msbuild found comes from the VS installation which _does_ have C++ installed.

I'm concerned though about pinning too precise of a path here based on specific toolset versions. This seems like it will silently cause the test to start skipping when Azure Pipelines updates from the 14.2 VC toolset to something later. I'd be more comfortable with a test that won't stop running where it currently runs without raising a fuss. Can we perhaps instead use `#if NETCORE` to skip the test so it always runs on .NET Framework?

qmfrederik:
Sure, no problem.

PS - I've noticed CI doesn't get triggered automatically on a push to a PR. Is that something that can be fixed?

AArnott:
> I've noticed CI doesn't get triggered automatically on a push to a PR. Is that something that can be fixed?

I sure hope so. [It's broken everywhere](https://developercommunity.visualstudio.com/content/problem/1213995/github-pr-triggers-does-not-work.html). Please consider voting that ticket up.

AArnott (approved):
Thanks.

src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
```diff
@@ -936,6 +940,7 @@ public async Task AssemblyInfo_SuppressedImplicitlyByTargetExt()
         Assert.Empty(result.LoggedEvents.OfType<BuildWarningEventArgs>());
     }
 
+#if !NETCORE
```
AArnott, line 943: This isn't the right condition. It should be `#if !NETCOREAPP`
As it is, the test ran on .NET Core anyway and failed the test run.

Issue #524, opened 2020-10-23
Consider narrowing the scope of GitExtensionsTest


qmfrederik:
Many unit tests in the `GitExtensionsTest` class are covering core functionality of NerdBank.GitVersioning, mainly via the `.GetHeight(...)` extension methods.

I believe the result of the version height calculation should be considered observable behavior of the `VersionOracle` class, and most of the unit tests in the `GitExtensions` class could be moved to the `VersionOracleTests` class and invoke the `VersionOracle` class directly.

For example, there's `GitExtensionsTest.GetVersionHeight_ProgressAndReset` and `VersionOracleTests.VersionHeightResetsWithVersionSpecChanges`. I think they are both testing essentially the same observable behavior, and these unit tests should be located in the same class.

This would help unit testing the changes in #521 .

@AArnott what do you think?

AArnott:
I tend to favor testing at the lowest level possible where public API makes sense. Now, maybe a lot of these `GitExtensions` shouldn't actually be *public*, but since they are, I would tend to exhaustively test them at that level, and then just have smoke tests at the higher level such as `VersionOracle` tests. 
It sounds like you're saying we are testing the same thing (at the same complexity level) at both places though, and that would sound unnecessarily redundant. However I'd want high confidence that it was truly redundant before I would delete or trim a test down. The coverage we have is important to maintain.

qmfrederik:
You raise a good point, though I think `GitExtensions` should not be public API (because it is closely coupled with LibGit2), and most of the methods in `GitExtensions` serve as helper methods which are used exclusively by unit tests.

Here's why. (I'll focus on the `GetVersionHeight` and `GetHeight` method for now, though most of the concerns apply to the other methods, too)

- Looking at the call sites for these methods (see below), only one 20-line overload is actually used by product code, by the `VersionOracle` class.
- All other overloads are only ever invoked by test code. In essence, they are shortcuts to enable testing of `GitWalkTracker`.
- If we're to make the switch to a managed Git implementation, `GitExtensions` _can't_ really be much of a public API, since _all_ of its methods are tightly coupled to the LibGit2 API.

To make sure we can re-use the unit tests which excercise `GitExtensions.GetVersionHeight` when we switch to a managed backend, I see two options:
- Rewrite the tests as unit tests create a new `VersionOracle` and validate the `VersionHeight` property.
[...]

PR #521, opened 2020-10-19
Add a managed Git implementation

qmfrederik:
Adds a managed Git implementation to NBGV. The managed implementation is used by default but libgit2sharp is used under a switch or when _writing_ to git is required (e.g. `nbgv prepare-release`). 
 
All tests run against both backends.

This implementation already contains a couple of performance optimizations which I worked on with the help of @filipnavara ; but the performance may have regressed as I focused on matching the NBGV features.

Perhaps worth mentioning is that:
- Opening files (i.e. `File.OpenRead`, but also `File.Exists` and just constructing the file name) turned out to be an expensive operation. On Windows, there's for example a P/Invoke call to CreateFile to implement some kind of `TryOpen`, to avoid having to call `File.Exists` followed by `File.OpenRead` (which is slow and racy)
- I did not know about git pack files before starting this work, and a fair amount of the code is related to parsing git pack files in a performant manner (e.g. pooling `FileStream` objects which access git pack files, and caching objects retrieved from a git pack in a cache, because this can be an expensive operation).
- For git pack files, `Streams` which point to the pack file are pooled instead of opening/closing them repeatedly (which is expensive).

Notable steps completed:
- [X] Add unit tests for the managed Git client
- [X] Review exceptions being thrown, make them more descriptive
- [x] `GitCommit`: Since most commits only have one/two parents, investigate whether it is worth storing the first (two) parents as a field on the `GitCommit` struct, and additional parents in a list. This would avoid allocating many `List<GitObjectId>` objects with only a single item.
- [x] `GitCommit`: Investigate whether it's worth parsing the `Author` value for the current HEAD only.
- [X] Add documentation for the managed Git client
- [x] Add support for Git alternates
- [x] ~~Investigate whether walking the Git tree can be parallelized~~ (deferred)
- [x] Test against [very large pack files](https://github.com/libgit2/libgit2sharp/issues/1835).
- [x] Test with shallow clone and fail with a better exception message than libgit2sharp fails with (an NRE)
- [x] ~~Test that .ps1 scripts in nuget package still work~~ (these scripts were removed as they are incompatible now)
- [x] Implement `ManagedGit.GitRepository.ShortenObjectId`
- [x] Implement `GitRepository.Lookup`

Closes #505
Closes #487
Closes #433
Closes #92

Improves perf as tracked by #114

PR #527, opened 2020-10-27
Re-route calls to GitExtensions.GitIdAsVersion, GetVersionHeight to VersionOracle

qmfrederik:
- Update unit tests in `GitExtensionsTests` which excercise `GitIdAsVersion` to test `VersionOracle.Version` instead
- Update unit tests in `GitExtensionsTests` which excercise `GetVersionHeight` to test `VersionOracle.VersionHeight` instead

As per discussion in #524, a follow-up PR can move these tests to the `VersionOracleTests` class.

Contributes toward #505, #521 by making it easier to unit-test a managed implementation of the `VersionOracle` class.

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/VersionFileTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/VersionFileTests.cs
@@ -485,10 +485,9 @@ public class VersionFileTests : RepoTestBase
     }
 
     [Theory]
-    [InlineData(false, false)]
-    [InlineData(true, false)]
-    [InlineData(true, true)]
-    public void VersionJson_Inheritance(bool commitInSourceControl, bool bareRepo)
+    [InlineData(false)]
+    [InlineData(true)]
+    public void VersionJson_Inheritance(bool commitInSourceControl)
     {
         if (commitInSourceControl)
         {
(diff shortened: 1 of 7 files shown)


src/NerdBank.GitVersioning.Tests/VersionFileTests.cs
```diff
@@ -485,10 +485,9 @@ public void VersionJson_DoNotInheritButNoVersionSpecified()
     }
 
     [Theory]
-    [InlineData(false, false)]
-    [InlineData(true, false)]
-    [InlineData(true, true)]
-    public void VersionJson_Inheritance(bool commitInSourceControl, bool bareRepo)
+    [InlineData(false)]
+    [InlineData(true)]
+    public void VersionJson_Inheritance(bool commitInSourceControl)
```
qmfrederik, line 490: `VersionOracle` has no support for bare repositories (how would you run `dotnet build`/`nbgv` on a repo which is not checked out?), so I think it's acceptable to not support this scenario.

AArnott: I just tested it, and you're right: `nbgv` doesn't work on bare repos, but that's something we *might* want to add support for later. While you can't _build_ such a repo, you could certainly use `nbgv get-version` and `nbgv get-commits` commands in one, if it worked.

AArnott (approved):
Thanks

Issue #540, opened 2020-11-27
Sporadic MSBuild hangs with managed git + .NET 5.0


qmfrederik:
I'm seeing sporadic hangs in MSBuild when using Nerdbank.GitVersioning 3.4.142-alpha with .NET SDK 5.0.100:

```
Microsoft (R) Build Engine version 16.8.0+126527ff1 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

Attempting to cancel the build...
/Users/vagrant/.nuget/packages/nerdbank.gitversioning/3.4.142-alpha/build/Nerdbank.GitVersioning.Inner.targets(17,5): warning MSB4220: Waiting for the currently executing task "Nerdbank.GitVersioning.Tasks.GetBuildVersion" to cancel.
/Users/vagrant/.nuget/packages/nerdbank.gitversioning/3.4.142-alpha/build/Nerdbank.GitVersioning.Inner.targets(17,5): warning MSB4220: Waiting for the currently executing task "Nerdbank.GitVersioning.Tasks.GetBuildVersion" to cancel.
/Users/vagrant/.nuget/packages/nerdbank.gitversioning/3.4.142-alpha/build/Nerdbank.GitVersioning.Inner.targets(17,5): warning MSB4220: Waiting for the currently executing task "Nerdbank.GitVersioning.Tasks.GetBuildVersion" to cancel.
/Users/vagrant/.nuget/packages/nerdbank.gitversioning/3.4.142-alpha/build/Nerdbank.GitVersioning.Inner.targets(17,5): warning MSB4220: Waiting for the currently executing task "Nerdbank.GitVersioning.Tasks.GetBuildVersion" to cancel.
```

I've observed the hang on macOS and Windows, I have not (yet) observed it on Linux.

The issue appears to be specific to .NET 5.0; I have not seen it when using .NET SDK 3.1.404.

I didn't find the time to dig deeper yet, but thought I'd be worth sharing the info nonetheless.

AArnott:
Thanks. Next time you see this, can you please capture a heap dump and share (privately)?

BenjaminAbt:
Same problem (also .NET SDK 5.0.100) happens to me on GitHub Action (Ubuntu Latest Pool) and Azure DevOps (Windows 2019 Pool) Pipeline but not locally.

```yaml
      - name: Versioning
        uses: dotnet/nbgv@master
        id: nbgv

      - name: 'Install .NET SDK'
[...]

PR #542, opened 2020-12-02
GitPackMemoryCache: Don't have multiple callers reuse the same stream

qmfrederik:
This appears to be a fix for #540. We must somehow end up with the same `GitPackMemoryCacheStream` stream being reused in an overlapped manner, causing multiple call sites to call `.Read`/`.Seek` at the same time, interfering with each other.

The naïve fix is to disable caching alltogether. I'm opening a PR so CI kicks in and we can get some perf numbers.

### DIFF
--- a/src/NerdBank.GitVersioning/ManagedGit/GitPackMemoryCacheViewStream.cs
+++ b/src/NerdBank.GitVersioning/ManagedGit/GitPackMemoryCacheViewStream.cs
@@ -1,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Nerdbank.GitVersioning.ManagedGit
+{
+    internal class GitPackMemoryCacheViewStream : Stream
+    {
+        private readonly GitPackMemoryCacheStream baseStream;
+
+        public GitPackMemoryCacheViewStream(GitPackMemoryCacheStream baseStream)
+        {
+            this.baseStream = baseStream ?? throw new ArgumentNullException(nameof(baseStream));
+        }
+
+        public override bool CanRead => true;
+
+        public override bool CanSeek => true;
+
+        public override bool CanWrite => false;
+
+        public override long Length => this.baseStream.Length;
+
+        private long position;
+
+        public override long Position
+        {
+            get => this.position;
+            set => throw new NotSupportedException();
+        }
+
+        public override void Flush() => throw new NotImplementedException();
+
+        public override int Read(byte[] buffer, int offset, int count)
+        {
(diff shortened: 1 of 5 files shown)


qmfrederik:
Yeah, I thought so. Re-using streams which have been 'de-deltafied' is a very important perf building block. Let me think about this a bit more.

qmfrederik:
Yeah, looks like this is the proper fix then.

In summary:
- Git pack files may contain deltafied streams. A single stream can be deltafied using a base stream which in itself is also deltafied.
- De-delatifying streams can be very expensive. If multiple streams rely on the same same deltafied stream _S_, it is _very_ efficient to cache stream _S_. Enter the `GitPackMemoryCache` and `GitPackMemoryCacheStream`.
- The same holds true for compressed streams; if a GitPack file contains a compressed stream which is 'hot' (accessed repeatedly), it is worth caching the decompressed stream.
- We somehow end up with multiple call sites accessing a `GitPackMemoryCacheStream` at the same time, issuing overlapped `Seek` and `Read` calls.
- The fix I came up with is to use a `GitPackMemoryCacheViewStream` (yeah, naming) which provides a view over a base `GitPackMemoryCacheStream`. This appears to fix the issue and keep perf consistent.

src/NerdBank.GitVersioning/ManagedGit/GitPackMemoryCacheViewStream.cs
```diff
@@ -0,0 +1,77 @@
...
+#if NETSTANDARD
+        public int Read(Span<byte> buffer)
+#else
+        /// <inheritdoc/>
+        public override int Read(Span<byte> buffer)
+#endif
+        {
+            int read = 0;
+
+            lock (this.baseStream)
```
AArnott, line 49: Where is concurrency introduced? I don't recall anywhere in NB.GV before or after your change that forks to another thread.

qmfrederik: I don't think threading is involved (as in, it's not a race condition), so the lock is probably not necessary. I'd need to dig a bit deeper to find out what exactly is happening - one theory is that a deltafied stream is using _itself_ as the base stream.

AArnott: If it is using itself, the `lock` won't prevent that. Given this PR fixes a serious issue, I'll merge. But I don't feel great about a `lock` that _should_ be a no-op. If your understanding of the issue improves over time and you get a chance, I'd welcome another PR.

Issue #547, opened 2020-12-05
Intermitent out-of-memory errors with the 3.4.153-alpha release in Visual Studio


Numpsy:
Hi,

I tried installing the 3.4.142-alpha release into a small project a while back and it all seemed ok, but then more recently tried in with a much larger solution and started seeing #540.

I've now tried updating that larger solution to the 3.4.153-alpha package, and that seems to work ok for a few builds, but then starts failing with

```
1>C:\Users\Richard Webb\.nuget\packages\nerdbank.gitversioning\3.4.153-alpha\build\Nerdbank.GitVersioning.Inner.targets(17,5): error MSB4018: The "Nerdbank.GitVersioning.Tasks.GetBuildVersion" task failed unexpectedly.
1>C:\Users\Richard Webb\.nuget\packages\nerdbank.gitversioning\3.4.153-alpha\build\Nerdbank.GitVersioning.Inner.targets(17,5): error MSB4018: System.IO.IOException: Not enough memory resources are available to process this command.
1>C:\Users\Richard Webb\.nuget\packages\nerdbank.gitversioning\3.4.153-alpha\build\Nerdbank.GitVersioning.Inner.targets(17,5): error MSB4018:
1>C:\Users\Richard Webb\.nuget\packages\nerdbank.gitversioning\3.4.153-alpha\build\Nerdbank.GitVersioning.Inner.targets(17,5): error MSB4018:    at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
1>C:\Users\Richard Webb\.nuget\packages\nerdbank.gitversioning\3.4.153-alpha\build\Nerdbank.GitVersioning.Inner.targets(17,5): error MSB4018:    at System.IO.MemoryMappedFiles.MemoryMappedView.CreateView(SafeMemoryMappedFileHandle memMappedFileHandle, MemoryMappedFileAccess access, Int64 offset, Int64 size)
1>C:\Users\Richard Webb\.nuget\packages\nerdbank.gitversioning\3.4.153-alpha\build\Nerdbank.GitVersioning.Inner.targets(17,5): error MSB4018:    at System.IO.MemoryMappedFiles.MemoryMappedFile.CreateViewAccessor(Int64 offset, Int64 size, MemoryMappedFileAccess access)
[...]

PR #548, opened 2020-12-07
Managed Git: .Dispose() housekeeping

qmfrederik:
The root cause of #547 is that `MemoryMappedStream` calls `SafeMemoryMappedViewHandle.AcquirePointer(ref this.ptr)` without calling `SafeMemoryMappedViewHandle.ReleasePointer()`, never releasing the (virtual) memory.

On top of that, `ManagedGitContext.Dispose()` did not call `GitRepository.Dispose()`, and the `GitPackMemoryCache` was missing a `Dispose` method which frees up all the streams stored in the cache.

With this PR, a loop like this in `VersionOracleManagedTests`:

```
for (int i = 0; i < int.MaxValue; i++)
{
    var oracle = this.GetVersionOracle("path_to_repository");
    var version = oracle.Version;
}
```

goes from leaking 1 GB of memory / minute to a fairly stable state.

### DIFF
--- a/src/NerdBank.GitVersioning/Managed/ManagedGitContext.cs
+++ b/src/NerdBank.GitVersioning/Managed/ManagedGitContext.cs
@@ -122,7 +122,12 @@ namespace Nerdbank.GitVersioning.Managed
         /// <inheritdoc/>
         protected override void Dispose(bool disposing)
         {
-            this.Repository.Dispose();
+            if (disposing)
+            {
+                this.Repository.Dispose();
+            }
+
+            base.Dispose(disposing);
         }
 
         /// <summary>
--- a/src/NerdBank.GitVersioning/Managed/ManagedGitContext.cs
+++ b/src/NerdBank.GitVersioning/Managed/ManagedGitContext.cs
@@ -119,6 +119,12 @@ namespace Nerdbank.GitVersioning.Managed
             return this.Repository.ShortenObjectId(this.Commit.Value.Sha, minLength);
         }
 
+        /// <inheritdoc/>
+        protected override void Dispose(bool disposing)
+        {
+            this.Repository.Dispose();
+        }
+
         /// <summary>
         /// Encodes a commit from history in a <see cref="Version"/>
         /// so that the original commit can be found later.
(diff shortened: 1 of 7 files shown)


src/NerdBank.GitVersioning/Managed/ManagedGitContext.cs
```diff
@@ -119,6 +119,12 @@ public override string GetShortUniqueCommitId(int minLength)
             return this.Repository.ShortenObjectId(this.Commit.Value.Sha, minLength);
         }
 
+        /// <inheritdoc/>
+        protected override void Dispose(bool disposing)
+        {
+            this.Repository.Dispose();
```
AArnott: Unless you call `base.Dispose(disposing);` after your own disposal logic, you block the base class from disposing its own stuff. At the moment it seems the base class's method for this is empty, but one can't count on that always being true.

AArnott: It's also important to only dereference other managed objects within `if (disposing)` to avoid crashing when an AppDomain is being shutdown and managed objects may be torn down out of order.

src/NerdBank.GitVersioning/ManagedGit/GitPackCache.cs
```diff
@@ -51,5 +52,10 @@ public abstract class GitPackCache
         /// A <see cref="Stream"/> which represents the cached entry.
         /// </returns>
         public abstract Stream Add(long offset, Stream stream);
+
+        /// <inheritdoc/>
+        public virtual void Dispose()
```
AArnott: I've revised this to conform to the .NET disposal pattern.

src/NerdBank.GitVersioning/ManagedGit/MemoryMappedStream.cs
```diff
@@ -124,5 +140,12 @@ public override void Write(byte[] buffer, int offset, int count)
         {
             throw new NotSupportedException();
         }
+
+        /// <inheritdoc/>
+        protected override void Dispose(bool disposing)
+        {
```
AArnott, line 146: I modified this to follow the disposal pattern by checking `disposing` first and invoking the base method.

AArnott (approved):
Thank you! 
I pushed a commit to fix some minor dispose pattern differences.

qmfrederik:
Thanks, @AArnott . Do you plan on pushing a package with this fix to NuGet?

AArnott:
Yes. already queued as 3.4.161-alpha

PR #554, opened 2020-12-22
Use git "commondir" for reference lookups

### LINKED ISSUE #553: WorkTree With managed Git
When runing `nbgv get-version` (with nbgv version v3.4.161-alpha+a44c1fae2a) in my git worktree i get the following error

    ERROR: The commit 0000000000000000000000000000000000000000 was not found in this repository.

When falling back to LibGit2 it works as expected

AArnott:
Fixes #553

### DIFF
--- a/src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs
+++ b/src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs
@@ -33,9 +33,7 @@ namespace Nerdbank.GitVersioning.ManagedGit
         /// <summary>
         /// Creates a new instance of the <see cref="GitRepository"/> class.
         /// </summary>
-        /// <param name="workingDirectory">
-        /// The current working directory. This can be a subdirectory of the Git repository.
-        /// </param>
+        /// <param name="workingDirectory"><inheritdoc cref="GitRepository(string, string, string, string)" path="/param[@name='workingDirectory']" /></param>
         /// <returns>
         /// A <see cref="GitRepository"/> which represents the git repository, or <see langword="null"/>
         /// if no git repository was found.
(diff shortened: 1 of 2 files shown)


AArnott:
@qmfrederik Can you please review?

src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs
```diff
@@ -128,13 +128,13 @@ public static GitRepository Create(string workingDirectory, string gitDirectory,
         public GitRepository(string workingDirectory, string gitDirectory, string commonDirectory, string objectDirectory)
         {
             this.WorkingDirectory = workingDirectory ?? throw new ArgumentNullException(nameof(workingDirectory));
-            this.GitDirectory = gitDirectory ?? throw new ArgumentNullException(nameof(gitDirectory));
+            this.GitHeadDirectory = gitDirectory ?? throw new ArgumentNullException(nameof(gitDirectory));
```
qmfrederik, line 131: I tried to keep the variable names consistent with the names used by Git. For example, the Git documentation references this directory as `$GIT_DIR` (see https://git-scm.com/docs/gitrepository-layout). So it may make sense to keep `GitDirectory` as the name

At the same time, I have to admit I also get lost in those names from time to time, and `GitHeadDirectory` does make the intent clear.

If you change the name to `GitHeadDirectory`, then these remarks:
- Perhaps the `Git` prefix (other variables are named `CommonDirectory` and `ObjectDirectory`, for example)
- Update the name of the `gitDirectory` argument in the constructor (and other methods that call the constructor).

qmfrederik:
Thanks, @AArnott. I have a minor remark regarding the variable names - up to you.

src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs
AArnott: Ya, I am conflicted on it too. But the desire to avoid bugs in this area won out for me. But the additional name changes you suggest makes sense. I'll apply those.

AArnott: Interesting. In reading the doc you linked to, I guess more than HEAD can arrear under this directory and mentions "a few known exceptions" that it unfortunately does not list. So I guess I'll leave the naming as it was but do my best to prominently comment to avoid bugs.

PR #575, opened 2021-03-29
Updated Cake.Core

azchohfi:


AArnott (approved):
Thanks for contributing.

pascalberger:
@AArnott Note that 1.x release have a compatible API and therefore suggestion is to reference lowest API compatible Cake version from addins, which would be 1.0 currently. See https://cakebuild.net/docs/extending/addins/best-practices#cake-reference

AArnott:
Makes sense. I'd take another PR if you want to fix it. I have no experience with Cake.

PR #586, opened 2021-04-12
Add NBGV_GitEngine=Disabled option

### LINKED ISSUE #568: Add an MSBuild property for disabling any NBGV tasks
### TL;DR 

It would be great to have something like

```bash
> dotnet build -p:NBGV_Disabled=true
```

### Longer version

I'm constantly hitting issues with NBGV in various inner-loop (developer machine, devcontainers, omnisharp, git submodule) scenarios. It usually works great once it's set up and running correctly in the CI environment, but I find myself adding a condition to the package import so I can turn it of when I don't need it, e.g.

```xml
<PackageReference Include="Nerdbank.GitVersioning" Version="..." Condition=" '$(NBGV_Disabled)' != 'true' " />
```

... so I can opt out of it via an environment variable. I usually don't care what the package version turns out to be while working on a project in my IDE. It would be great if something like this would be supported out of the box.

m0sa:
This adds `Disabled` as an option to the `NBGV_GitEngine` setting, which turns off all git operations, but still emits expected / dummy AssemblyInfo properties.

Closes #568


src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
```diff
@@ -55,6 +55,22 @@ protected override void ApplyGlobalProperties(IDictionary<string, string> global
         => globalProperties["NBGV_GitEngine"] = "LibGit2";
 }
 
+[Trait("Engine", "Disabled")]
```
m0sa, line 58: @AArnott not sure if this is the best way to test this, but seemed like the obvious choice to get started.

rcdailey:
I can't tell easily from the code, since I'm unfamiliar with the architecture. Will this also address usages of `MyAssembly` when it is disabled?

src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
AArnott: That seems reasonable.

AArnott (commented):
This looks good. Not what I think I would have done but I really like where you took it. Let's run with it. I did a couple touch-ups and modified the test file slightly. I think we need to promote more tests to run in the no-git scenario than the two I did so far. Can you find the other tests that would pass or should pass, and add a couple new ones specifically for the no-git scenario?

AArnott:
I've resolved conflicts locally and will push once I validate the result.

Issue #595, opened 2021-04-26
GIT alternates not supported with version 3.4x


Sam13:
When repository is checked out with GIT alternates (e.g. as TeamCity does) NBGV get-version fails:
`Nerdbank.GitVersioning.GitException: An commit object with SHA 163380e3fa79d36c0e51efc251132aa7e7f86b9d could not be found.`

To reproduce:
_git clone --mirror https://github.com/dotnet/Nerdbank.GitVersioning.git_
_git clone --shared --reference .\Nerdbank.GitVersioning.git https://github.com/dotnet/Nerdbank.GitVersioning.git test_
_cd test_
_nbgv Get-Version_

I'm using NBGV v3.4.194 on Windows.
Setting _NBGV_GitEngine=LibGit2_ works but unfortunately not in Docker image _mcr.microsoft.com/dotnet/sdk:5.0_ because of dependency load problems for LibGit2.

AArnott:
It looks like yet another storage variance with git that our new managed implementation may not yet support?
@qmfrederik what do you think?

qmfrederik:
There was some work done to support alternates, e.g. https://github.com/dotnet/Nerdbank.GitVersioning/blob/master/src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs#L537-L568 (and in the constructor), but obviously it fails here.

I'll try to free up some time to have a look at the open managed-git issues (this one seems easy to reproduce). (Unfortunately, freeing up time is not so easy lately).

AArnott:
> Unfortunately, freeing up time is not so easy lately

Understandable. I _really_ appreciate how responsive you've been and how well you're supporting your new feature. We don't hear a lot about folks unblocked in 3.4 on various linux distros as a result of your work, but I assume that's because people mostly give feedback when something's broken rather than when it works. And I used to get a stream of complaints about various distros, and I haven't heard a peep since your work, so that's great! 
@Sam13 even mentioned the libgit2 workaround doesn't work for him because he is in a linux environment where it doesn't work, so clearly your new feature is having a positive impact.

qmfrederik:
Thanks, @AArnott . I had a look at two issues which seemed straightforward (including this one). Issues like #584 and #587 will probably need more time. Do you think you could label issues related to the managed git implementation with a managed-git label or something similar? That would help me identify those.

qmfrederik:
[...]

PR #596, opened 2021-04-27
Improve alternate parsing

qmfrederik:
There are at least two bugs in alternate parsing:
- On all platforms, the last value in the alternate files is ignored
- On Windows, the colon in full paths (e.g. `C:\Users\Me`) is used as a separator

This PR attempts to fix both issues. Contributes towards #595

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/ManagedGit/GitRepositoryTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/ManagedGit/GitRepositoryTests.cs
@@ -286,6 +286,25 @@ namespace ManagedGit
             }
         }
 
+        [Fact]
+        public void ParseAlternates_SingleValue_Test()
+        {
+            var alternates = GitRepository.ParseAlternates(Encoding.UTF8.GetBytes("/home/git/nbgv/.git/objects\n"));
+            Assert.Collection(
+                alternates,
+                a => Assert.Equal("/home/git/nbgv/.git/objects", a));
+        }
+
+        [Fact]
+        public void ParseAlternates_TwoValues_Test()
+        {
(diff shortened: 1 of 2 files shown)


src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs
```diff
@@ -714,5 +713,33 @@ public static unsafe string GetString(ReadOnlySpan<byte> bytes)
...
         }
+
+        /// <summary>
+        /// Parses the contents of the alternates file, and returns a list of (relative) paths to the alternate object directories.
+        /// </summary>
+        /// <param name="alternates">
+        /// The contents of the alternates files.
+        /// </param>
+        /// <returns>
+        /// A list of (relative) paths to the alternate object directories.</returns>
```
AArnott, line 724:
```suggestion
        /// A list of (relative) paths to the alternate object directories.
        /// </returns>
```

src/NerdBank.GitVersioning.Tests/ManagedGit/GitRepositoryTests.cs
```diff
@@ -286,6 +286,25 @@ public void GetMissingObjectByShaTest()
             }
         }
 
+        [Fact]
```
AArnott, line 289: Can we add a test for your `c:/files` scenario?

AArnott (approved):
Looks good. A couple quick things to consider.

AArnott:
@qmfrederik If you push to this PR again, keep in mind I rebased your commits to allow for targeting the fix to v3.4. You should do a fetch and `git reset --hard origin/features/alternates` before doing more work to avoid inadvertently merging `master` into `v3.4`.

Issue #583, opened 2021-04-08
FileNotFoundException with managed git implementation


ssa3512:
One of our developers is reporting an exception trying to build after latest updates to NBGV. After setting `NBGV_GitEngine` environment variable to `Libgit2` the error went away. He indicated the file mentioned `'C:\Users\dbarrett\source\repos\isolved\.git\objects\pack\pack-00314245d0ca77c6c10a5f87655b8b1e37bbac25.pack'.` does in fact not exist but this doesn't seem to cause any issues with Libgit2.

```
Build started...
1>------ Build started: Project: AOS.Core, Configuration: Debug Any CPU ------
NuGet package restore failed. Please see Error List window for detailed warnings and errors.
Error occurred while restoring NuGet packages: The operation failed as details for project AOS.PayrollOffice.WebUI.Foundation could not be loaded.
1>C:\Users\dbarrett\.nuget\packages\nerdbank.gitversioning\3.4.194\build\Nerdbank.GitVersioning.Inner.targets(17,5): error MSB4018: The "Nerdbank.GitVersioning.Tasks.GetBuildVersion" task failed unexpectedly.
1>C:\Users\dbarrett\.nuget\packages\nerdbank.gitversioning\3.4.194\build\Nerdbank.GitVersioning.Inner.targets(17,5): error MSB4018: System.IO.FileNotFoundException: Could not find file 'C:\Users\dbarrett\source\repos\isolved\.git\objects\pack\pack-00314245d0ca77c6c10a5f87655b8b1e37bbac25.pack'.
1>C:\Users\dbarrett\.nuget\packages\nerdbank.gitversioning\3.4.194\build\Nerdbank.GitVersioning.Inner.targets(17,5): error MSB4018: File name: 'C:\Users\dbarrett\source\repos\isolved\.git\objects\pack\pack-00314245d0ca77c6c10a5f87655b8b1e37bbac25.pack'
1>C:\Users\dbarrett\.nuget\packages\nerdbank.gitversioning\3.4.194\build\Nerdbank.GitVersioning.Inner.targets(17,5): error MSB4018:    at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
[...]

PR #597, opened 2021-04-27
Don't load git packs if the .pack file is missing

qmfrederik:
Should contribute towards #583


AArnott (approved):
Thanks.

AArnott:
@qmfrederik  After this, should we ask the original reporter of the issue to test, or do you think more work is required?

qmfrederik:
@AArnott I think we can ask @ssa3512 for confirmation (but he may have already deleted the orphaned index file), other than that, I don't see much else we can do.

PR #599, opened 2021-05-01
Abort prepare-release command when version on current branch is not incremented

### LINKED ISSUE #591: Weird behaviour when nextVersion is equal current version
Maybe I'm just doing wrong, but this behaviour looks weird to me:

From development branch which has `"version": "2.0-dev.{height}"` and run `nbgv prepare-release --nextVersion 2.2 rc` the development branch changes correctly to `"version": "2.2-dev.{height}"`, and release branch to `"version": "2.0-rc.{height}"`

However, when (unintentionally) setting next version to the current version like `run nbgv prepare-release --nextVersion 2.0 rc`, the cli output is (kind of) correct
```
release/v2.0 branch now tracks v2.0-rc.{height} stabilization and release.
development branch now tracks v2.0-dev.{height} development.
```
but in reality it changes the prerelease tag in development branch also to "**rc**" so both development and release branch now have `"version": "2.0-rc.{height}"`.
Aside that it should have warned me about not increasing the version, but I would not have expected the prerelease tag in development to be touched!

In fact there seems to be some checking already when the nextversion is set to something lower (ie going from 2.1 to 2.0):
`"Cannot change version from 2.1-dev.{height} to 2.0-dev.{height} because 2.0-dev.{height} is older than 2.1-dev.{height}."`, so it may need to check if it's lower or same version in that case.

ap0llo:
When the next version for the current branch is set by the user using the `--nextVersion` command line parameter, the next version might be equal to the current version.
In that case no commit is created (no changes on the branch), which in turn causes the current branch to be fast-forwarded to the commit on the newly-created release branch when the branches are merged.

To prevent this error, abort command if both versions match (there is already a check in place that prevents the version from being decremented).

The check only applies to the current branch because the version on the release branch not being incremented is a valid use case.

This should fix #591

### DIFF
--- a/src/NerdBank.GitVersioning/ReleaseManager.cs
+++ b/src/NerdBank.GitVersioning/ReleaseManager.cs
@@ -48,6 +48,11 @@
             VersionDecrement,
 
             /// <summary>
+            /// Branch cannot be set to the specified version because the new version is not higher than the current version
+            /// </summary>
+            NoVersionIncrement,
+
+            /// <summary>
             /// Cannot create a branch because it already exists
             /// </summary>
             BranchAlreadyExists,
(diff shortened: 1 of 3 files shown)


src/NerdBank.GitVersioning/ReleaseManager.cs
```diff
@@ -273,6 +278,14 @@ public void PrepareRelease(string projectDirectory, string releaseUnstableTag =
 
             var nextDevVersion = this.GetNextDevVersion(versionOptions, nextVersion, versionIncrement);
 
+            // check if the current version on the main branch is different from the next version
```
AArnott, line 281:
```suggestion
            // check if the current version on the current branch is different from the next version
```

AArnott (approved):
Looks great. Thanks for the added tests as well.

Issue #587, opened 2021-04-12
v3.4 regressed version height calculation involving path filters


marcominerva:
Hi!

I have a project with this _version.json_ file:

```
{
	"$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
	"version": "1.2",
	"publicReleaseRefSpec": [
		"^refs/heads/master$" // we release out of master
	],
	"nugetPackageVersion": {
		"semVer": 2
	},
	"pathFilters": [ "." ]
}
```

With this configuration, If I use Nerdbank.GitVersioning version 3.3.37, everything works like a charm. However, as soon as I update it to the v3.4.194, the git versioning isn't calculated anymore (it is always set to 0).

AArnott:
Thanks for reporting. We have had a few regressions in 3.4 due to switching from libgit2 to our own managed git implementation. See our [release notes](https://github.com/dotnet/Nerdbank.GitVersioning/releases/tag/v3.4.142-alpha) for how to switch 3.4 back to libgit2 as a potential workaround.

I haven't heard of the symptom you describe. Can you provide repro steps that we can use?

marcominerva:
You can see the problem in this repository: https://github.com/marcominerva/TinyHelpers/tree/develop/src/TinyHelpers. Now the Nerdbank.GitVersioning is version 3.3.37, so when I build the project (that contains the _version.json_ file I shown you above), the version is correctly set to 1.2.xxx. If, instead, I update the version to v3.4.194 and just retry to rebuild the project, the version returns to the 1.2.0.

AArnott:
Great. I can repro the problem. The workaround I linked you to is effective as well. I'll continue investigating.

AArnott:
@qmfrederik this regression is apparently because some code was never converted for the managed git implementation around path filters support:

https://github.com/dotnet/Nerdbank.GitVersioning/blob/8126d95e2c789efb42f9a89cf576fe75e69119fc/src/NerdBank.GitVersioning/Managed/ManagedGitExtensions.cs#L197-L216

Do you recall why this wasn't done? Is it going to be something we can fill in?

qmfrederik:
[...]

PR #606, opened 2021-05-25
Path filters: Recurse into directories if child entries may be included by filters

qmfrederik:
`GitExtensions.IsRelevantCommit` will determine whether a commit is relevant when calculating version height. A commit is considered relevant if any of the files which are changed in that commit match the path filters.

This code relies on `FilterPath.Includes` to determine whether a specific path is included in by a path filter.

The unmanaged implementation gets a list of full file names, whereas the managed implementation recurses the directory tree. 

This causes a bug where the managed implementation would not recurse into directories whose child items may be included.
For example, if you have a path filter to include `src/lib`, `FilterPath.Includes("src")` would return `false`, and changes to any files in `src/lib/` would not be considered.

This should fix the root cause of #587.

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/VersionOracleTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/VersionOracleTests.cs
@@ -732,6 +732,28 @@ public abstract class VersionOracleTests : RepoTestBase
     }
 
     [Fact]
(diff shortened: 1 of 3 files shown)


qmfrederik:
@AArnott I updated the unit test you created to use `/` as a path separator instead of `\`.

AArnott (approved):
Thanks. I rebased your commits to get them so they can target v3.4 instead of master.

PR #624, opened 2021-06-23
Informational version supports four components

rcdailey:
If the user specifies a three-component version number in their
`version.json`, for example `"1.2.3"`, then the InformationalVersion
property will show something like `"1.2.3.50+metadata"`, where `50` is
the build height.

The previous behavior is still retained when users specify a
two-component version number, such as `"1.2"`.

### DIFF
--- a/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
+++ b/src/NerdBank.GitVersioning.Tests/BuildIntegrationTests.cs
@@ -1050,8 +1050,14 @@ public abstract class BuildIntegrationTests : RepoTestBase, IClassFixture<MSBuil
 
         string expectedBuildMetadataWithoutCommitId = additionalBuildMetadata.Any() ? $"+{string.Join(".", additionalBuildMetadata)}" : string.Empty;
 
+        var optionalFourthComponent = versionOptions.VersionHeightPosition switch
+        {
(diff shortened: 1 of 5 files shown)


rcdailey:
@AArnott Hey sorry to ping. Just wondering if/when we can get around to getting this merged. Is there more I can assist with on this PR?

I'm fully expecting that the way I've coded this may not be acceptable. I'm not familiar with the architecture, so I've likely mixed concerns where it is not appropriate. Happy to make any adjustments necessary.

AArnott:
This seems like a very reasonable goal. I'm surprised it's been this long before this enhancement was made. Thank you for your submission and your patience in my getting to it. Your tests and changes were great. I just made a couple small stylistic changes and will merge shortly.

rcdailey:
Andrew, I know you stay busy so I tried not to pester you too much about it. But you've been extremely helpful in chat with my questions over the last few months, and very accepting of my contribution here. Thank you for making FOSS enjoyable!

epvanhouten:
@AArnott  Can we get a v3.5 or a v3.5-alpha with this change in it? Any issues that need to be resolved elsewhere to help that happen?

AArnott:
I'm releasing it now. Look for 3.5.50-alpha on nuget.org to appear within an hour.

PR #634, opened 2021-07-28
Fix bugs with > 2 GB pack files

filipnavara:
Unfortunately it's non-trivial to make a test for this since it requires really large data. The obvious bug is the incorrect mask in the `.idx` file reading which resulted in incorrect offset being read. The other two bugs happen if a delta is crossing the 2GB boundary in the pack file.

### DIFF
--- a/src/NerdBank.GitVersioning/ManagedGit/GitPackIndexMappedReader.cs
+++ b/src/NerdBank.GitVersioning/ManagedGit/GitPackIndexMappedReader.cs
@@ -128,7 +128,7 @@ namespace Nerdbank.GitVersioning.ManagedGit
             {
                 // If the first bit of the offset address is set, the offset is stored as a 64-bit value in the table of 8-byte offset entries,
                 // which follows the table of 4-byte offset entries: "large offsets are encoded as an index into the next table with the msbit set."
-                offset = offset & 0x7FF;
+                offset = offset & 0x7FFFFFFF;
 
                 offsetBuffer = this.Value.Slice(offsetTableStart + 4 * objectCount + 8 * (int)offset, 8);
                 var offset64 = BinaryPrimitives.ReadInt64BigEndian(offsetBuffer);
--- a/src/NerdBank.GitVersioning/ManagedGit/GitPackReader.cs
+++ b/src/NerdBank.GitVersioning/ManagedGit/GitPackReader.cs
@@ -44,7 +44,7 @@ namespace Nerdbank.GitVersioning.ManagedGit
             if (type == GitPackObjectType.OBJ_OFS_DELTA)
             {
                 var baseObjectRelativeOffset = ReadVariableLengthInteger(stream);
-                var baseObjectOffset = (int)(offset - baseObjectRelativeOffset);
+                var baseObjectOffset = offset - baseObjectRelativeOffset;
 
                 var deltaStream = new ZLibStream(stream, decompressedSize);
                 var baseObjectStream = pack.GetObject(baseObjectOffset, objectType);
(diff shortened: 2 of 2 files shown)


src/NerdBank.GitVersioning/ManagedGit/GitPackReader.cs
```diff
@@ -44,7 +44,7 @@ public static Stream GetObject(GitPack pack, Stream stream, long offset, string
             if (type == GitPackObjectType.OBJ_OFS_DELTA)
             {
                 var baseObjectRelativeOffset = ReadVariableLengthInteger(stream);
-                var baseObjectOffset = (int)(offset - baseObjectRelativeOffset);
+                var baseObjectOffset = offset - baseObjectRelativeOffset;
```
filipnavara, line 47: Example: Offset is `4 000 000 000`, relative offset is `300 000`. The result would be incorrectly truncated into wrong value.

src/NerdBank.GitVersioning/ManagedGit/GitPackIndexMappedReader.cs
```diff
@@ -128,7 +128,7 @@ public override (long?, GitObjectId?) GetOffset(Span<byte> objectName, bool ends
             {
                 // If the first bit of the offset address is set, the offset is stored as a 64-bit value in the table of 8-byte offset entries,
                 // which follows the table of 4-byte offset entries: "large offsets are encoded as an index into the next table with the msbit set."
-                offset = offset & 0x7FF;
+                offset = offset & 0x7FFFFFFF;
```
filipnavara, line 131: Ref: https://github.com/go-git/go-git/blob/master/plumbing/format/idxfile/idxfile.go#L146
Ref: https://git-scm.com/docs/pack-format

> If the large offset chunk exists and the 31st bit is on, then removing that bit reveals the row in the large offsets containing the 8-byte offset of this object.

src/NerdBank.GitVersioning/ManagedGit/GitPackReader.cs
```diff
@@ -98,9 +98,9 @@ private static (GitPackObjectType, long) ReadObjectHeader(Stream stream)
             return (type, length);
         }
 
-        private static int ReadVariableLengthInteger(Stream stream)
+        private static long ReadVariableLengthInteger(Stream stream)
         {
-            int offset = -1;
+            long offset = -1;
```
filipnavara, line 103: Ref: https://git-scm.com/docs/pack-format

> Observation: length of each object is encoded in a variable length format and is not constrained to 32-bit or anything.

Example: Base offset is `4 000 000 000`, delta relative offset would be `3 000 000 000`. The computed delta offset would thus be `1 000 000 000`. The number `3 000 000 000` must fit in the data type returned here.

src/NerdBank.GitVersioning/ManagedGit/GitPackReader.cs
AArnott: nit: the type is no longer obvious here. Since we had a bug in the type, it seems useful to make it explicit. Perhaps you can make L46-L47 both explicit. For example:

```suggestion
                long baseObjectOffset = offset - baseObjectRelativeOffset;
```

AArnott (approved):
Thanks for identifying the bug and especially for offering the fix!
@qmfrederik any other thoughts?

qmfrederik:
Looks good to me, thanks for fixing this!

filipnavara:
Thanks for reviews! It was the last blocker for the managed GIT backend on our huge and unusual repository (where even libgit2 sometimes struggles).

AArnott:
This is shipping to nuget.org as 3.4.231 and should appear in about an hour.

filipnavara:
> This is shipping to nuget.org as [3.4.231](https://github.com/dotnet/Nerdbank.GitVersioning/releases/tag/v3.4.231).

Thanks! We finally updated internally to the managed GIT implementation. It was a long journey but I am really happy about it.

PR #636, opened 2021-07-28
Profile benchmarks

qmfrederik:
This PR attempts to add profiling to the benchmark runs. Let's see how it goes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -196,3 +196,5 @@ FakesAssemblies/
 GeneratedArtifacts/
 _Pvt_Extensions/
 ModelManifest.xml
+
+BenchmarkDotNet.Artifacts/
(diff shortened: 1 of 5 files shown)


AArnott:
It looks like you may be close. :)

As profiling impacts the runtime perf, and we generally just want to measure perf, ideally this change can be under a pipeline parameter so we can check the box to profile a build where without that it just does the normal thing.

qmfrederik:
Looks like we're seeing https://github.com/dotnet/BenchmarkDotNet/issues/1731, so I used a preview version of BenchmarkDotNet. The AppVeyor package feed doesn't produce signed packages, though, so I had to disable package signature validation.

Let me see if I can get profiling to work at all, and worry about the NuGet configuration later 😄 .

qmfrederik:
So it looks like this actually gets some profiling done and creates SpeedScope files. But https://speedscope.app/ barks at them:

> Unrecognized format! See documentation about supported formats.

Thoughts?

AArnott:
I've never heard of speedscope files, so I'm afraid I can't offer help there. But it sounds like progress. :)

PR #648, opened 2021-08-27
Fixes build with msbuild `-graph -isolate` switches

### LINKED ISSUE #647: Build fails when using `msbuild -graph -isolate`
When building with the `-graph -isolate` switches, the build fails when Nerdbank.GitVersioning invokes the MSBuild task, for which no `ProjectReference` item exists.

See also [MSBuild static graph](https://github.com/dotnet/msbuild/blob/main/documentation/specs/static-graph.md#what-is-static-graph).

AArnott:
Fixes #647 

- [x] Document breaking changes around when `BuildMetadata` items have to be declared.
- [x] File bugs against msbuild for how the global property syntax of P2Ps is interpreted differently by static graph than the msbuild task itself.

### DIFF
--- a/src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.Inner.targets
+++ b/src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.Inner.targets
@@ -42,5 +42,11 @@
     <Message Importance="low" Text="AssemblyInformationalVersion: $(AssemblyInformationalVersion)" />
     <Message Importance="low" Text="NuGetPackageVersion: $(NuGetPackageVersion)" />
   </Target>
+  
+  <!-- These targets are called by MSBuild static graph because we declare a ProjectReference item to this file. -->
+  <Target Name="GetTargetFrameworks" />
+  <Target Name="GetNativeManifest" />
+  <Target Name="GetCopyToOutputDirectoryItems" />
+  <Target Name="GetTargetFrameworksWithPlatformForSingleTargetFramework" />
 
 </Project>
--- a/src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
+++ b/src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
@@ -76,6 +76,8 @@
   <!-- Compile a list of global properties that may vary when a project builds but that would never influence the result of the GetBuildVersion task. -->
   <ItemGroup>
     <NBGV_GlobalPropertiesToRemove Include="TargetFramework" />
+    <NBGV_GlobalPropertiesToRemove Include="Configuration" />
+    <NBGV_GlobalPropertiesToRemove Include="Platform" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
+++ b/src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
@@ -79,6 +79,8 @@
     <NBGV_GlobalPropertiesToRemove Include="RuntimeIdentifier" />
     <NBGV_GlobalPropertiesToRemove Include="Configuration" />
     <NBGV_GlobalPropertiesToRemove Include="Platform" />
+  
+    <_BuildMetadataSnapped Include="@(BuildMetadata)" />
   </ItemGroup>
 
   <ItemGroup>
@@ -97,6 +99,8 @@
   </ItemGroup>
 
   <Target Name="GetBuildVersion" Returns="$(BuildVersion)">
+    <Error Text="BuildMetadata items changed after a copy was made. Add all BuildMetadata items before importing this file." Condition=" '@(BuildMetadata)' != '@(_BuildMetadataSnapped)' " />
+    
     <!-- Calculate version by invoking another "project" with global properties that will serve as a key 
          into an msbuild cache to ensure we only invoke the GetBuildVersion task as many times as will produce a unique value. -->
     <MSBuild Projects="@(ProjectReference)"
--- a/src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
+++ b/src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
@@ -91,13 +91,18 @@
       <BuildReference>false</BuildReference>
       <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
       <Visible>false</Visible>
+      <NBGV_InnerProject>true</NBGV_InnerProject>
     </ProjectReference>
   </ItemGroup>
 
   <Target Name="GetBuildVersion" Returns="$(BuildVersion)">
     <!-- Calculate version by invoking another "project" with global properties that will serve as a key 
          into an msbuild cache to ensure we only invoke the GetBuildVersion task as many times as will produce a unique value. -->
-    <MSBuild Projects="$(NBGV_CoreTargets)" Properties="$(NBGV_InnerGlobalProperties)" RemoveProperties="@(NBGV_GlobalPropertiesToRemove)" Targets="GetBuildVersion_Properties">
+    <MSBuild Projects="@(ProjectReference)"
+             Condition=" '%(ProjectReference.NBGV_InnerProject)' == 'true' "
+             Properties="%(ProjectReference.Properties)"
+             RemoveProperties="%(ProjectReference.GlobalPropertiesToRemove)"
+             Targets="GetBuildVersion_Properties">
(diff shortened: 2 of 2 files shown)


src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.Inner.targets
```diff
@@ -42,5 +42,11 @@
     <Message Importance="low" Text="AssemblyInformationalVersion: $(AssemblyInformationalVersion)" />
     <Message Importance="low" Text="NuGetPackageVersion: $(NuGetPackageVersion)" />
   </Target>
+  
+  <!-- These targets are called by MSBuild static graph because we declare a ProjectReference item to this file. -->
+  <Target Name="GetTargetFrameworks" />
+  <Target Name="GetNativeManifest" />
+  <Target Name="GetCopyToOutputDirectoryItems" />
+  <Target Name="GetTargetFrameworksWithPlatformForSingleTargetFramework" />
```
AArnott, line 50: The build broke until I defined these targets, although they do nothing. It seems they were invoked directly by the engine rather than the common .targets.
@rainersigwald Shouldn't the msbuild static graph tolerate missing targets rather than make me define these?

src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
```diff
@@ -62,31 +62,48 @@
...
+    <NBGV_InnerGlobalProperties Condition=" '$(GitRepoRoot)' != '' ">$(NBGV_InnerGlobalProperties)GitRepoRoot=$(GitRepoRoot);</NBGV_InnerGlobalProperties>
+    <NBGV_InnerGlobalProperties Condition=" '$(PublicRelease)' != '' ">$(NBGV_InnerGlobalProperties)PublicRelease=$(PublicRelease);</NBGV_InnerGlobalProperties>
+    <NBGV_InnerGlobalProperties Condition=" '$(_NBGV_BuildingRef)' != '' ">$(NBGV_InnerGlobalProperties)_NBGV_BuildingRef=$(_NBGV_BuildingRef);</NBGV_InnerGlobalProperties>
+    <NBGV_InnerGlobalProperties Condition=" '$(ProjectPathRelativeToGitRepoRoot)' != '' ">$(NBGV_InnerGlobalProperties)ProjectPathRelativeToGitRepoRoot=$(ProjectPathRelativeToGitRepoRoot);</NBGV_InnerGlobalProperties>
+    <NBGV_InnerGlobalProperties Condition=" '$(GitVersionBaseDirectory)' != '' ">$(NBGV_InnerGlobalProperties)ProjectDirectory=$(GitVersionBaseDirectory);</NBGV_InnerGlobalProperties>
+    <NBGV_InnerGlobalProperties Condition=" '$(OverrideBuildNumberOffset)' != '' ">$(NBGV_InnerGlobalProperties)OverrideBuildNumberOffset=$(OverrideBuildNumberOffset);</NBGV_InnerGlobalProperties>
```
AArnott, line 72: I had to break out every global property as defined in the removed lines below with this, with each one conditioned on being non-null. Apparently although the MSBuild task is fine to accept empty-valued global properties, the static graph's `ProjectReference.Properties` parser is *not*.
@rainersigwald Is this a bug in msbuild?

src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
```diff
@@ -62,31 +62,52 @@
...
+      <!-- Do our very best to prevent Microsoft.Common.CurrentVersion.targets or IDEs from processing this P2P. It's only here for MSBuild's static graph. -->
+      <BuildReference>false</BuildReference>
+      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+      <Visible>false</Visible>
+      <NBGV_InnerProject>true</NBGV_InnerProject>
+    </ProjectReference>
+  </ItemGroup>
+
+  <Target Name="GetBuildVersion" Returns="$(BuildVersion)">
+    <Error Text="BuildMetadata items changed after a copy was made. Add all BuildMetadata items before importing this file." Condition=" '@(BuildMetadata)' != '@(_BuildMetadataSnapped)' " />
```
AArnott, line 102: This represents a takeback in Nerdbank.GitVersioning that was necessitated by the static graph.
In particular, the global properties used to be assembled, based in part on `@(BuildMetadata)`, as part of the msbuild target. Now that I have to express all global properties at evaluation time, the `@(BuildMetadata)` items have to be complete *before* this file is imported. I had to make a fix to one place of my own in another repo to get the BuildMetadata moved up. 
This error is here to assist with folks upgrading past this change so that if they would regress, they won't do it silently.
Strictly speaking this requirement is true of all properties as well. But as msbuild properties are all set in evaluation before any items are, they're safer by default. Only if those properties were set by an msbuild target are they at risk of regressing.
FYI: @rainersigwald

src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
```diff
@@ -62,31 +62,52 @@
...
+    <ProjectReference Include="$(NBGV_CoreTargets)">
+      <Targets>GetBuildVersion_Properties;GetBuildVersion_CloudBuildVersionVars</Targets>
+      <Properties>$(NBGV_InnerGlobalProperties)BuildMetadata=@(BuildMetadata, ',');</Properties>
+      <GlobalPropertiesToRemove>@(NBGV_GlobalPropertiesToRemove)</GlobalPropertiesToRemove>
 
+      <!-- Do our very best to prevent Microsoft.Common.CurrentVersion.targets or IDEs from processing this P2P. It's only here for MSBuild's static graph. -->
+      <BuildReference>false</BuildReference>
+      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+      <Visible>false</Visible>
+      <NBGV_InnerProject>true</NBGV_InnerProject>
```
AArnott, line 97: @rainersigwald Can you think of any more metadata I can or should set to stay off the radar of other P2P processors?

src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.Inner.targets
rainersigwald: This is tracked by dotnet/msbuild#4252.

src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
rainersigwald: This one I think is probably just How It Is for isolate mode. dotnet/msbuild#3877 may be related but the whole idea is that we need to know what would get built up front.

src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
rainersigwald: I'm trying to decide whether it's better to set `SkipGetTargetFrameworkProperties=true` or (as you've done) define an empty `GetTargetFrameworks`. I lean toward the former but don't feel strongly.

It doesn't appear that we have an explicit "reference this project in this custom way" operation, and I think we should. dotnet/msbuild#6852.

src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.Inner.targets
AArnott: I see you essentially voted against that issue though. Is that how you still feel?

src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
AArnott: @rainersigwald  I tried that change, but the graph failed because it still expects `GetTargetFrameworks` to be defined:
https://github.com/dotnet/Nerdbank.GitVersioning/compare/msbuildgraphisolate...SkipGetTargetFrameworkProperties

AArnott: I've added the additional metadata, but left the empty target in place so that the scenario doesn't break.

rainersigwald: Oh, and I see why: we don't condition the target map on anything (and it applies to all references). Answers that question (define the empty GetTargetFrameworks).

src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.Inner.targets
rainersigwald: No, updated there to indicate that I think it's reasonable now.

PR #652, opened 2021-09-10
inheritance fixes and better path filtering

akobr:
- inheritance of version options is more clear, and the process is more controllable
- added clear way how to control the inheritance of path filters (property )
  - `inheritPathFilters`: A value indicating whether path filters are inherited from parent version configurations. This is a backward compatibility flag, if you want the behaviour of inheritance on path filters, you must set the flag to true explicitly. 
- support for hierarchical versioning based on the folder structure
  - `hierarchicalVersion`: A value indicating whether the version in the current directory structure is hierarchical and calculated independently per each sub-directory. When this option has been enabled, the functionality of path filters is ignored, and the version of each folder is affected only by its own content.
  - When this flag is true, the path filters are not applied, and rather each sub-folder automatically applies itself as an only filter (included).

### DIFF
--- a/src/NerdBank.GitVersioning/FilterPath.cs
+++ b/src/NerdBank.GitVersioning/FilterPath.cs
@@ -10,7 +10,7 @@ namespace Nerdbank.GitVersioning
     /// <summary>
     /// A filter (include or exclude) representing a repo relative path.
     /// </summary>
-    public class FilterPath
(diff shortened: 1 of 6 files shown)


akobr:
I'm so sorry. Unfortunately, I opened this PR here instead of in my fork repository. 😊

PR #653, opened 2021-09-11
Add Cake alias for the "cloud" command

ap0llo:
This PR adds the Cake alias `GitVersioningCloud()` which exposes the functionality of `nbgv cloud` to Cake build scripts.

To share the implementation between the Cake and nbgv versions, I moved the bulk of the implementation into the `CloudCommand` class in the main `Nerdbank.GitVersioning` project.

There are some aspects for which I'm unsure if I found the right solution. Please let me know if I should change them.

- There might be a bette name than `CloudCommand` .
- I added a test specifc to `Cake.GitVersioning` but put it into `NerdBank.GitVersioning.Tests`. Perhaps adding a separate test project `Cake.GitVersioning.Tests` would be a better solution
- The existing Cake alias had this code snippet in it:
   ```cs
      string directoryName = Path.GetDirectoryName(Assembly.GetAssembly(typeof(GitVersioningAliases)).Location);

      if (string.IsNullOrWhiteSpace(directoryName))
      {
            throw new InvalidOperationException("Could not locate the Cake.GitVersioning library");
      }
   ```
   I don't understand why this is necessary but added it to the new alias as well.


src/Cake.GitVersioning/GitVersioningAliases.cs
```diff
@@ -45,5 +41,46 @@ public static VersionOracle GitVersioningGetVersion(this ICakeContext cakeContex
...
+        [CakeMethodAlias]
+        public static void GitVersioningCloud(this ICakeContext cakeContext, string projectDirectory = ".", GitVersioningCloudSettings settings = null)
+        {
+            var fullProjectDirectory = (new DirectoryInfo(projectDirectory)).FullName;
+
+            string directoryName = Path.GetDirectoryName(Assembly.GetAssembly(typeof(GitVersioningAliases)).Location);
+
+            if (string.IsNullOrWhiteSpace(directoryName))
+            {
+                throw new InvalidOperationException("Could not locate the Cake.GitVersioning library");
```
AArnott, line 68: I'm curious when this would *ever* happen. When does `GetDirectoryName` return the empty string given a non-empty input?

src/Cake.GitVersioning/GitVersioningCloudSettings.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>
+        /// Defines a few common version variables as cloud build variables, with a "Git" prefix.
+        /// </summary>
+        public bool CommonVariables { get; set; } = false;
+
+        /// <summary>
+        /// Additional cloud build variables to define.
+        /// </summary>
+        public IDictionary<string, string> AdditionalVariables { get; set; } = new Dictionary<string, string>();
```
AArnott: How important is it that these collection-typed properties be settable? I want this dictionary to use case-insensitive keys, but if the caller can replace the dictionary with their own, then a fix I make here may not even stick. In cake, is setting these properties important?

src/NerdBank.GitVersioning.Tests/GitVersioningCloudProviderTests.cs
```diff
@@ -0,0 +1,30 @@
...
+    [Fact]
+    public void HasExpectedValues()
+    {
+        var expectedValues = CloudBuild.SupportedCloudBuilds.Select(cb => cb.GetType().Name);
+        var actualValues = Enum.GetNames(typeof(GitVersioningCloudProvider));
+
+        var missingValues = expectedValues.Except(actualValues);
+        Assert.True(
+            !missingValues.Any(),
+            $"Enumeration is missing the following values of supported cloud build providers: {string.Join(", ", missingValues)}"
```
AArnott: Nice! 💯

AArnott (approved):
Thank you.

src/Cake.GitVersioning/GitVersioningAliases.cs
ap0llo: I assumed this was related to some cake behavior I'm unaware of, because this was also in the existing `GitVersioningGetVersion` cake alias, 
Since the code is in the `Cake.GitVersioning` library, I don't see why it needs to check if its own assembly has a parent directory.

AArnott: Thanks. I left it in just in case.

PR #662, opened 2021-09-24
Use nearest relevant commit id based on path filters (fixes #643)

### LINKED ISSUE #643: Behavior of pathFilter and commit ID
Hi,
we are using pathFilter to increment versions only for affected components in a monorepo structure. This is working fine regarding the height but the behavior of commit ID for non-public releases is different to what (at least) I expected.

Given a monorepo with /componentA and /componentB subdirectories and according pathFilter configurations. Assuming a commit in /componentA-subfolder:

- Current behavior
  - height+1 and new commit ID for componentA (expected)
  - height+0 for componentB (expected), but also new commit ID (not expected)
- Expected behavior
  - commit ID only changes if the commit falls in the range of the pathFilter

Does this make sense or am I missing something?

tg73:
This is work in progress for discussion. I've also been affected by #643 so I'm attempting a fix.

@AArnott I'd welcome comments on the general approach. I've implemented in managed only for now, plan to implement for libgit if/when the managed implementation is acceptable.

The basic idea is to leverage the work of `GitWalkTracker`, keeping track of the commit with greatest height (which I believe should be the relevant commit nearest to the 'current' commit, and only commits that survive path filtering are tracked).

Note the comment [here](https://github.com/zanyants/Nerdbank.GitVersioning/blob/7a99e3c06e3530df0b0e8d41a2d0fff6567501e6/src/NerdBank.GitVersioning/Managed/ManagedGitExtensions.cs#L49) - I don't have a deep enough understanding of this scenario to know what should be done.

### DIFF
--- a/src/NerdBank.GitVersioning/GitContext.cs
+++ b/src/NerdBank.GitVersioning/GitContext.cs
@@ -176,7 +176,7 @@ namespace Nerdbank.GitVersioning
         /// <returns>A string that is at least <paramref name="minLength"/> in length but may be more as required to uniquely identify the git object identified by <see cref="GitCommitId"/>.</returns>
         public abstract string GetShortUniqueCommitId(int minLength);
 
-        internal abstract int CalculateVersionHeight(VersionOptions? committedVersion, VersionOptions? workingVersion);
+        internal abstract (int height, string? nearestRelevantCommit) CalculateVersionHeightAndNearestRelevantCommit(VersionOptions? committedVersion, VersionOptions? workingVersion);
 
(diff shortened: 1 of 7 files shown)


tg73:
@AArnott  I'm waiting for feedback on this before proceeding. Some libgit2-based tests currently fail because they intentionally throw `NotImplementedException` at this time. Tools and tasks using the managed git implementation work (at least for me) and demonstrate the intended fix, and the code changes show the principle of operation. Please let me know if you require anything further.

hayer:
Has this been implemented?

AArnott:
Thank you for submitting this. Sorry it's taken so long to look at it. Given the conflicts and age, I created a fresh PR (#1447), which supersedes this one.

PR #686, opened 2021-11-11
GitHubActions cloud build exception on local machine

rcdailey:
If I run the following command on my development machine (outside of a Github Actions Workflow):

    nbgv cloud -s GitHubActions -a -p src

I get the following exception:

> Unhandled exception: System.Reflection.TargetInvocationException:
> Exception has been thrown by the target of an invocation.
> ---> System.ArgumentNullException: Value cannot be null. (Parameter 'path')
>   at System.IO.File.AppendAllText(String path, String contents)

The `EnvironmentFile` variable is null because `GITHUB_ENV` does not
exist on a local machine. Instead, simply write to console so the user
can see the output of the cloud build variables (which I believe is the
intention when a developer runs this locally).

### DIFF
--- a/src/NerdBank.GitVersioning/CloudBuildServices/GitHubActions.cs
+++ b/src/NerdBank.GitVersioning/CloudBuildServices/GitHubActions.cs
@@ -28,7 +28,16 @@
 
         public IReadOnlyDictionary<string, string> SetCloudBuildVariable(string name, string value, TextWriter stdout, TextWriter stderr)
         {
-            Utilities.FileOperationWithRetry(() => File.AppendAllText(EnvironmentFile, $"{Environment.NewLine}{name}={value}{Environment.NewLine}"));
+            if (EnvironmentFile is not null)
+            {
+                Utilities.FileOperationWithRetry(() =>
+                    File.AppendAllText(EnvironmentFile, $"{Environment.NewLine}{name}={value}{Environment.NewLine}"));
+            }
+            else
+            {
+                (stdout ?? Console.Out).WriteLine($"{name}={value}");
+            }
+
             return GetDictionaryFor(name, value);
         }
 


src/NerdBank.GitVersioning/CloudBuildServices/GitHubActions.cs
```diff
@@ -28,7 +28,16 @@ internal class GitHubActions : ICloudBuild
...
         {
-            Utilities.FileOperationWithRetry(() => File.AppendAllText(EnvironmentFile, $"{Environment.NewLine}{name}={value}{Environment.NewLine}"));
+            if (EnvironmentFile is not null)
+            {
+                Utilities.FileOperationWithRetry(() =>
+                    File.AppendAllText(EnvironmentFile, $"{Environment.NewLine}{name}={value}{Environment.NewLine}"));
+            }
+            else
+            {
+                (stdout ?? Console.Out).WriteLine($"{name}={value}");
```
AArnott, line 38: We do this for other cloud providers where it's truly a logging command. But for github it is not. As it is not natural to write to the console, we should not 'silently' redirect the logging from a file to the console as if there were no problem.
At the _very_ least, we would want to write to Console.Error and emit a warning that the environment variable is missing and then emit what we would have written to the file.

But I'm more inclined to just emit that error and throw. If someone wants to see what would have happened, they can set the environment variable to a file path and then inspect the file.
The error we print about the missing env var could give the name of the env var so that users could know what to set in order to see it work.

AArnott (changes requested):

src/NerdBank.GitVersioning/CloudBuildServices/GitHubActions.cs
rcdailey: Maybe I'm misunderstanding the purpose of running `nbgv cloud` locally, but to me it's what I use to understand what kinds of variables will be available in my cloud CI pipeline. Sadly it doesn't even accomplish that, as the actual syntax for accessing cloud build variables is completely different. For example in Github Actions it's `steps.nbgv.outputs.PrereleaseVersion` not `NBGV_PrereleaseVersion`. But at least it's something.

I always found it how to use cloud variables very nondeterministic, mostly due to lack of documentation. And also the fact that syntax is vastly different between providers. Maybe a long term goal should be to address this root concern, either through documentation or a better "preview" subcommand.

I don't really like your suggestion of setting an environment variable to get this working. It takes away from the simplicity of the command (Right now I can just run it real quick to see what it spits out). I can do this without setting environment variables with VisualStudioTeamServices. I don't really like that it has to be different, much less harder, on different cloud platforms.

Again it could be I'm using the wrong tool for the job.

AArnott: `nbgv cloud` is *not* intended for local invocation. It's meant to be included in your cloud build script. It sets the cloud build variables and/or cloud build number of that build. That's why it writes whatever commands, in whatever way, as required by the cloud build system that it auto-detects.

The `nbgv cloud -h` command tells you which variables will be set when you specify `-c`. And if you want them _all_ set using `-a`, you can get the full list by running `nbgv get-version -f json`. Does that help?

rcdailey: When I run `nbgv cloud -h` I do not see variables in the output.

AArnott:
```
cloud:
  Communicates with the ambient cloud build to set the build number and/or other cloud build variables.

Usage:
  nbgv cloud [options]

Options:
  -p, --project <project>                                     The path to the project or project directory used to
                                                              calculate the version. The default is the current
                                                              directory. Ignored if the -v option is specified.
  --metadata <metadata>                                       Adds an identifier to the build metadata part of a
                                                              semantic version. [default: System.String[]]
  -v, --version <version>                                     The string to use for the cloud build number. If not
                                                              specified, the computed version will be used.
  -s, --ci-system                                             Force activation for a particular CI system. If not
  <AppVeyor|AtlassianBamboo|GitHubActions|GitLab|Jenkins|T    specified, auto-detection will be used. Supported values
  eamCity|Travis|VisualStudioTeamServices>                    are: AppVeyor, VisualStudioTeamServices, GitHubActions,
                                                              TeamCity, AtlassianBamboo, Jenkins, GitLab, Travis
  -a, --all-vars                                              Defines ALL version variables as cloud build variables,
                                                              with a "NBGV_" prefix.
  -c, --common-vars                                           Defines a few common version variables as cloud build
                                                              variables, with a "Git" prefix (e.g. GitBuildVersion,
                                                              GitBuildVersionSimple, GitAssemblyInformationalVersion).
  -d, --define <define>                                       Additional cloud build variables to define. Each should
                                                              be in the NAME=VALUE syntax. [default: System.String[]]
  -?, -h, --help                                              Show help and usage information
```

zooming in:
```
  -c, --common-vars                                           Defines a few common version variables as cloud build
                                                              variables, with a "Git" prefix (e.g. GitBuildVersion,
                                                              GitBuildVersionSimple, GitAssemblyInformationalVersion).
```

zooming in further:

> GitBuildVersion, GitBuildVersionSimple, GitAssemblyInformationalVersion

That's what you get for `-c`, and for `-a`, you get it as I explained in my last post.

rcdailey: Thanks, it's a bit buried in there. I was looking for a line-by-line list, I wasn't sure what to look for. Thanks for pointing it out.

PR #689, opened 2021-11-13
Update to .NET 6, Update libgit2sharp to include macOS ARM64

drasticactions:
In order to use GitVersioning with the .NET 6 SDK on the M1, we need to update libgit2sharp to the newest release, which includes macOS ARM64 (https://github.com/libgit2/libgit2sharp.nativebinaries/pull/132) support.

So, in order to use that, we also need to update the base library to include .NET 6. 

I also had to include some using fixes for the tests, since it got an ambiguous reference between 'System.IO.Compression.ZLibStream' and 'Nerdbank.GitVersioning.ManagedGit.ZLibStream'.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -36,7 +36,7 @@ stages:
       matrix:
         linux:
           imageName: 'ubuntu-20.04'
-          testModifier:
+          testModifier: -f net6.0
         windows:
           imageName: 'windows-2019'
           testModifier:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -36,7 +36,7 @@ stages:
       matrix:
         linux:
           imageName: 'ubuntu-20.04'
-          testModifier: -f net6.0
+          testModifier:
         windows:
           imageName: 'windows-2019'
           testModifier:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -56,10 +56,10 @@ stages:
       displayName: Configure git commit author for testing
 
     - task: UseDotNet@2
-      displayName: Install .NET Core 6.0.100 SDK
+      displayName: Install .NET Core 3.1 runtime
       inputs:
-        packageType: sdk
-        version: 6.0.100
+        packageType: runtime
+        version: 3.1.x
 
     - task: UseDotNet@2
       displayName: Install .NET Core 5.0.401 SDK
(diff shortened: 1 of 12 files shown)


AArnott (approved):
Looks great! Thank you for contributing.

drasticactions:
Sorry, I had to fix the source generator, I didn't know you would commit on top of this while I was trying to fix it, lol.

It should actually build now!

drasticactions:
Gahhh, alright, I got as close as I could. I'm not sure what needs to be flipped for the x86 tests on Windows to run and why it's now getting BadImageExceptions. The other tests should be okay, and they all run locally on my M1 Max Mac and pass. 

I can come back to fix that test, but if you want to, you can commit to it. I won't disrupt, lol.

Thanks again for building this!

AArnott:
BadImageFormatException typically means you're loading the wrong an assembly with a mismatched bitness. For example the x86 test process may be trying to load an x64 binary. That binary could by libgit2 or it could even be an arch-specific managed assembly. 

I may have time to look into it today.

AArnott:
I fixed a couple more issues, but the kicker is that the MSBuild.Locator is finding the 64-bit MSBuild when running on .NET 5. The odd thing is we've had .NET 5 in a 32-bit process find 32-bit MSBuild before, so I don't know what about this PR broke it. I'll keep looking into it. There may be a bug elsewhere that we need to workaround.

AArnott:
I'm tackling a smaller number of problems in #696 first. Then I'll come back to this.

AArnott:
@drasticactions now that my smaller PR has completed, which included a libgit2sharp upgrade, is there anything more you actually need? Your PR would add *testing* on net6.0, which is not something that my PR did.

drasticactions:
I think so? The only thing that might need to be added is this, https://github.com/dotnet/Nerdbank.GitVersioning/blob/a0e4dd7b20a8f4eaa18c2d8369879748d17a4f16/src/NerdBank.GitVersioning/LibGit2/LibGit2GitExtensions.cs#L176

There is a call to look for the correct libgit2 library, and that might need the call to get the ARM64 version, but I'm not sure if that code is called or not.

IMO you can close this PR and add that change yourself if it's needed.

AArnott:
Done in #697. Thanks for catching what would likely have been a regression from my change.

Issue #401, opened 2019-11-03
Support height in fourth version part for NuGet


clairernovotny:
While the majority of uses can be done with with a major.minor version and the third/patch as the auto increment, there is sometimes a need for a four part version.

If I use 1.2.3 as the version, the version shows as 1.2.3.1, but the NuGetPackageVersion shows as 1.2.3.

I believe this should be supported. I'd want 1.2.3.1 as the package version.

onovotny:
already works.

AArnott:
@bradwilson also wanted something similar: height in the 4th part of *file* version, but no 4th part in the nuget version.

bradwilson:
My usage is slightly different (I want height as part of the prerelease information for the NuGet package, so 1.2.3-pre.1+hash)

azeno:
@bradwilson I believe this was possible in version 2, see #477

ap0llo:
I have a use-case for this (the first 3 parts of the package version should be constant and I would just like to automatically increment the 4th part).

Would you accept a PR adding this?

I'd suggest to add a `precision` property to `nugetPackageVersion` in version.json (analogous to the `assemblyVersion` settings)

AArnott:
Sure. That seems like a reasonable design. Thanks for offering the PR.

ap0llo:
I implemented support for this in #719.

Thinking about it, would it also make sense to have something similar for the cloud build number so the build number could be configured to match the NuGet package version in the case a 4th version field is used?

AArnott:
Yes, we want cloud build numbers to be mostly unique as well, so if height has been moved to the 4th integer, the cloud build number should include that as well.

ap0llo:
I had an idea for an alternative way to solve this.

Instead of adding a `precision` setting, what do you think about automatically including the 4th version field in the NuGet package version if 3 fields are configured in `version.json`.
The same behavior could then be applies to the cloud build number.

ap0llo:
I experimented with the idea of always emitting a 4-part NuGet package version if there are 3 parts configured in `version.json` but I think that would be a breaking change (at least it breaks a lot of tests) so it would probably be best to go with the `precision` setting anyways.

Would that also be your preferred solution for the cloud build number?
I'd be happy to add that as well

AArnott:
I think cloud build number could go the way you suggest of automatically including the 4th integer if 3 are specified in version.json.
[...]

PR #723, opened 2022-02-17
Include all four version components in the cloud build number if three or more components are set in the version file

ap0llo:
As discussed in #401, this PR adapts calculation of the cloud build number to include all 4 version fields if either

- The commit is configured to appear in the cloud build number as 4th version component (existing behavior) or
- 3 version components are set in the version file and thus the git height is at the 4th position or
- All 4 version components are set in the version file


AArnott (approved):
Thank you!

PR #801, opened 2022-09-05
Caching git heights

AArnott:
As first seen in #499, which was inadvertently closed prior to merging. Recreating here for review and consideration.

## Summary
Adds caching of git heights. As calculating the height of repositories with large volumes of commits is expensive, caching the git heights can save time in the following circumstances:
- Repetitive invocations of the `GetBuildVersion` msbuild task- e.g. during `dotnet pack`'ing a project, the `GetBuildVersion` task is invoked four times
- Incremental versioning- in cases where a cached height is available for older commits, this value will be used avoid the cost of recalculating the entire git height (only new commits will need to be traversed)

### When can cached results be used?
- The base version to calculate the height for is the same as the cached base version
- The relative path of the cached height matches the requested relative path 
- The commit to calculate the height for is either the same OR a child of the cached commit id

### Opting out
The caching behavior is enabled by default but can be opted-out by setting the new `NerdbankGitVersioningUseHeightCache` msbuild property to `false`

### Cache file
A `version.cache.json` file is created per-project with contents that look like this:
```
/*Cached commit height, created by Nerdbank.GitVersioning. Do not modify.*/{"BaseVersion":"1.0","Height":1636,"CommitId":"f30bb9e26af1dc80f5c38f43d9d6a7b4f770bb14","RelativeProjectDir":""}
```

### Automated testing
- Verified `GitHeightCache` can serialize + deserialize heights correctly
- Verified caching has a measurable impact on performance for cases when there are many commits to traverse
- All existing tests pass

### Manual testing
Consumed locally-packed version of `Nerdbank.GitVersioning` in a C# project and added ~1500 commits, verifying:
- Height caching takes effect on second build, dramatically decreasing build time (~10s -> ~1s)
- Adding an additional single commit can leverage the cached version for the previous ~1500 commits but adds the latest
- Setting the `NerdbankGitVersioningUseHeightCache` property to `false` bypasses the caching behavior.


AArnott:
CC @djluck @tftomsun

src/NerdBank.GitVersioning/GitExtensions.cs
```diff
@@ -56,11 +60,33 @@ public static int GetVersionHeight(this Commit commit, string repoRelativeProjec
...
                 baseVersion != null ? SemanticVersion.Parse(baseVersion.ToString()) :
                 versionOptions.Version ?? SemVer0;
-
+            
             var versionHeightPosition = versionOptions.VersionHeightPosition;
+            
             if (versionHeightPosition.HasValue)
             {
-                int height = commit.GetHeight(repoRelativeProjectDirectory, c => CommitMatchesVersion(c, baseSemVer, versionHeightPosition.Value, tracker));
+                var cache = new GitHeightCache(commit.GetRepository().Info.WorkingDirectory, versionOptions.RelativeFilePath, baseVersion);
```
djluck, line 68: We should probably be doing this operation only when `useHeightCaching` is true.

djluck:
Hey @AArnott, apologies for the incredibly slow response- work was incredibly stressful for a period and the idea of finding the time to look through an MR I wrote a good year ago was just too much!

So full disclosure: it's been a long time since I've thought about this MR and may not be the most reliable reviewer :) I left a comment on one minor thing I think we could improve. I think we should also provide some updates to the documentation regarding this, informing people about:
- Why caching can be useful
- How to opt out
- What `version.cache.json` is to add it to their `.gitignore` files

AArnott:
Optimizations have been made since this PR that makes this one out-dated.

Issue #817, opened 2022-09-20
ManagedGit: Attempted to read past the end of the stream.


marcin-krystianc:
I've encountered an issue with managed git engine, which is related o packed streams but it is different from the https://github.com/dotnet/Nerdbank.GitVersioning/issues/797.

- repro: [TestNBGV-size0.zip](https://github.com/dotnet/Nerdbank.GitVersioning/files/9606609/TestNBGV-size0.zip) 
- error message:
```
The "Nerdbank.GitVersioning.Tasks.GetBuildVersion" task failed unexpectedly. [C:\Users\marcink\.nuget\packages\nerdbank.gitversioning\3.5.109\build\PrivateP2PCaching.proj]
System.InvalidOperationException: Unable to get version from commit: bfc6aaadccbbdf5facd2a539af89dbc1e9c975f9 [C:\Users\marcink\.nuget\packages\nerdbank.gitversioning\3.5.109\build\PrivateP2PCaching.proj]
 ---> System.IO.EndOfStreamException: Attempted to read past the end of the stream. [C:\Users\marcink\.nuget\packages\nerdbank.gitversioning\3.5.109\build\PrivateP2PCaching.proj]
   at Nerdbank.GitVersioning.ManagedGit.StreamExtensions.ReadAll(Stream stream, Span`1 buffer) [C:\Users\marcink\.nuget\packages\nerdbank.gitversioning\3.5.109\build\PrivateP2PCaching.proj]
   at Nerdbank.GitVersioning.ManagedGit.GitTreeStreamingReader.FindNode(Stream stream, ReadOnlySpan`1 name) [C:\Users\marcink\.nuget\packages\nerdbank.gitversioning\3.5.109\build\PrivateP2PCaching.proj]
   at Nerdbank.GitVersioning.ManagedGit.GitRepository.GetTreeEntry(GitObjectId treeId, ReadOnlySpan`1 nodeName) [C:\Users\marcink\.nuget\packages\nerdbank.gitversioning\3.5.109\build\PrivateP2PCaching.proj]
   at Nerdbank.GitVersioning.Managed.ManagedVersionFile.GetVersion(GitCommit commit, String repoRelativeProjectDirectory, Dictionary`2 blobVersionCache, String& actualDirectory) [C:\Users\marcink\.nuget\packages\nerd

[...]

PR #818, opened 2022-09-20
DeltaStreamReader should interpret instruction with size 0 as 0x10000

marcin-krystianc:
Fixes: https://github.com/dotnet/Nerdbank.GitVersioning/issues/817

https://git-scm.com/docs/pack-format/2.31.0#_instruction_to_copy_from_base_object:
"In its most compact form, this instruction only takes up one byte (0x80) with both offset and size omitted, which will have default values zero. There is another exception: size zero is automatically converted to 0x10000."


AArnott (approved):
Thank you for the analysis and the fix! 
I rebased your commit so we can service 3.5 with this fix rather than 3.6, which would keep this bug fix out of most people's hands for a while longer.

PR #826, opened 2022-09-25
Fix suboptimal incremental build due to faulty CompareFiles task

### LINKED ISSUE #825: IsContentOfFilesTheSame assumes full Stream.Read
In the  Nerdbank.GitVersioning.Tasks.CompareFiles class, the IsContentOfFilesTheSame method calls Stream.Read and assumes that a return value smaller than requested means that it got to the end of the file. As mentioned in <https://github.com/dotnet/Nerdbank.GitVersioning/issues/797#issuecomment-1251251500>, such an assumption is generally a bug. Here though, the stream refers to a regular file, which makes the assumption more likely to hold.

<https://github.com/dotnet/Nerdbank.GitVersioning/blob/719b0d50b809aece0f2c82d3ee27b2829dab4224/src/Nerdbank.GitVersioning.Tasks/CompareFiles.cs#L127-L133>

The CompareFiles task is used in Nerdbank.GitVersioning.targets, for deciding whether to run the Copy task. The bug might cause the task to consider the files different when they are actually identical; the target would then needlessly copy the file and cause MSBuild to execute other targets, resulting in a slower incremental build but eventually the correct artefacts.

It might be possible to delete the CompareFiles task altogether and use the SkipUnchangedFiles parameter of the Copy task instead.

<https://github.com/dotnet/Nerdbank.GitVersioning/blob/a908c6d9d2ed9b032d63ff885b4ccc97fc32046d/src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets#L144-L149>
<https://github.com/dotnet/Nerdbank.GitVersioning/blob/a908c6d9d2ed9b032d63ff885b4ccc97fc32046d/src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets#L181-L186>

AArnott:
The task was faulty because it assumed `Stream.Read` would fill the buffer if at all possible. 

Fixes #825

### DIFF
--- a/src/Nerdbank.GitVersioning.Tasks/CompareFiles.cs
+++ b/src/Nerdbank.GitVersioning.Tasks/CompareFiles.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using System.IO;
+    using System.IO.MemoryMappedFiles;
     using System.Linq;
     using System.Text;
     using Microsoft.Build.Framework;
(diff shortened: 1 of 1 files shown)


src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
```diff
@@ -143,10 +143,7 @@
                   />
     <!-- Avoid applying the newly generated AssemblyVersionInfo.cs file to the build
          unless it has changed in order to allow for incremental building. -->
-    <Nerdbank.GitVersioning.Tasks.CompareFiles OriginalItems="$(VersionSourceFile)" NewItems="$(NewVersionSourceFile)">
-      <Output TaskParameter="AreChanged" PropertyName="AssemblyVersionInfoChanged" />
-    </Nerdbank.GitVersioning.Tasks.CompareFiles>
-    <Copy Condition=" '$(AssemblyVersionInfoChanged)' == 'true' " SourceFiles="$(NewVersionSourceFile)" DestinationFiles="$(VersionSourceFile)" />
+    <Copy SourceFiles="$(NewVersionSourceFile)" DestinationFiles="$(VersionSourceFile)" SkipUnchangedFiles="true" />
```
AArnott, line 146: In testing, this parameter isn't working. The timestamp is being updated for every build anyway. :(

Issue #828, opened 2022-09-30
Versioning issue with MAUI


espenrl:
Using `Nerdbank.GitVersioning` together with MAUI is not that easy with regards to setting `ApplicationVersion` and `ApplicationDisplayVersion` properties. The main problem is setting the properties after `Nerdbank.GitVersioning` has exceuted, but before generating the manifests for Android, iOS and Windows.

The following code takes care of all this, and also adheres to the versioning requirements for each platform. The targets `FixVersionFor*` will make sure `SetApplicationVersion` is executed at the right time in the MSBuild pipeline.

NOTE: This code only takes care of Android, iOS and Windows.

``` xml
    <Target Name="FixVersionForAndroid"
            Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'"
            BeforeTargets="_GetAndroidPackageName"
            DependsOnTargets="SetApplicationVersion" />

    <Target Name="FixVersionForIOS" 
            Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'"
            BeforeTargets="_CompileAppManifest"
            DependsOnTargets="SetApplicationVersion" />

    <Target Name="FixVersionForWindows" 
            Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'" 
            BeforeTargets="MauiGeneratePackageAppxManifest" 
            DependsOnTargets="SetApplicationVersion" />

    <Target Name="SetApplicationVersion" DependsOnTargets="GetBuildVersion">

        <PropertyGroup>
            
            <_MajorStr>$([System.String]::Format("{0}", $([System.Version]::Parse("$(BuildVersion)").Major)))</_MajorStr>
            <_MinorStr>$([System.String]::Format("{0:D2}", $([System.Version]::Parse("$(BuildVersion)").Minor)))</_MinorStr>
            <_BuildStr>$([System.String]::Format("{0:D3}", $([System.Version]::Parse("$(BuildVersion)").Build)))</_BuildStr>

            <!-- Android requirement: ApplicationVersion must be a positive integer (used as an internal version number) -->
            <!-- iOS requirement: ApplicationVersion must be a three part version number -->
            <!-- Windows requirement: ApplicationVersion must be blank when ApplicationDisplayVersion is a four part version number -->
[...]

PR #860, opened 2022-11-09
Allow use of `NBGV_SetMauiVersionProperties` to suppress new behavior

AArnott:
Fixes issue @bakerhillpins brought up in https://github.com/dotnet/Nerdbank.GitVersioning/issues/828#issuecomment-1301098701

### DIFF
--- a/src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
+++ b/src/Nerdbank.GitVersioning.Tasks/build/Nerdbank.GitVersioning.targets
@@ -219,11 +219,11 @@
       <RazorCompile Include="$(VersionSourceFile)" />
     </ItemGroup>
   </Target>
-  
+
   <!-- Support for Maui projects -->
 
   <Target Name="NBGV_SetVersionForMauiAndroid"
-          Condition="'$(TargetPlatformIdentifier)' == 'android'"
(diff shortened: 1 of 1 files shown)


bakerhillpins:
@AArnott  This change allows the user to opt out of BOTH settings at once which is convenient when one is using central package management and they don't want to apply it to their EXE project.

It also appears that this feature is OFF by default as I don't see any change where `NBGV_SetMauiVersionProperties` is set to true.  Thus users would need to Opt IN to version updates?
Was your plan to opt in or out? My thought would be that users would expect oping out.

AArnott:
This is written to allow users to opt out. It remains on by default.
Note the check is that the new property _not_ be `false`. No one sets it to `false`, so the condition is satisfied and the behavior is _on_. Opting _out_ would look like this:

```xml
<NBGV_SetMauiVersionProperties>false</NBGV_SetMauiVersionProperties>
```

or

```
msbuild /p:NBGV_SetMauiVersionProperties=false
```

bakerhillpins:
Thanks for the clarification and sorry about the confusion.  I thought undefined properties evaluated false but I can't find the documentation concerning that. LOL

AArnott:
msbuild properties are not boolean. They are all strings. They default to evaluating to the empty string.

PR #876, opened 2022-12-15
Add git-based detection of tags at HEAD to improve PublicRelease detection

### LINKED ISSUE #873: How to rebuild past tagged builds as public releases? / Add BuildingRef override
My use case: I have a GitHub Actions cron job that checks if my docker base images have updated. If they have, it checks out the latest release tag according to GitHub Releases and rebuilds my Docker images.

My `version.json` looks like this:

```json
{
  "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
  "version": "0.7",
  "publicReleaseRefSpec": [
    "^refs/heads/release/v\\d+(?:\\.\\d+)?$",
    "^refs/tags/v\\d+\\.\\d+"
  ],
  "cloudBuild": {
    "buildNumber": {
      "enabled": true
    }
  },
  "release": {
    "branchName": "release/v{version}"
  }
}
```

As I'm checking out a different commit than the one chosen by GitHub (which defaults to the newest commit on the default branch for a cron job), the information in `GITHUB_REF` is stale. Now, I've seen #795 and adapted my workflow config accordingly:

(Assume a tag e.g. `v0.7.3` as `inputs.build-ref`)

```yml
jobs:
  build-and-push:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v3
      with:
        fetch-depth: 0 # avoid shallow clone so nbgv can do its work.
        ref: ${{ inputs.build-ref }}
    - uses: dotnet/nbgv@v0.4
      id: nbgv
      with:
        # see https://github.com/dotnet/nbgv/issues/50
        # remove this as soon as nbgv 3.6.x+ is released
        toolVersion: 3.6.65-alpha # or later
        toolFeed: https://pkgs.dev.azure.com/andrewarnott/OSS/_packaging/PublicCI/nuget/v3/index.json
      env:
        # if we checkout a build ref from input above, this is important
        IGNORE_GITHUB_REF: true # see https://github.com/dotnet/Nerdbank.GitVersioning/blob/274a42ed3321bcc43632dc970d65677f99a486a9/src/NerdBank.GitVersioning/CloudBuildServices/GitHubActions.cs#L32
```

As expected, the resulting `BuildingRef` does not point to the stale information from `GITHUB_REF` anymore, which is great.

However, nbgv also tells that I'm not building a public release now. I am, however, at a commit that is part of a `release/v0.7` branch and also at the commit where `v0.7.3` points. This is easy to reproduce locally:

`git checkout refs/tags/v0.7.3` results in a detached head state. Executing `nbgv get-version -f json` afterwards also shows a non-public release. Now I think it might make sense for detached heads to assume they are non-public. Also, looking at what I believe is the [relevant code](https://github.com/dotnet/Nerdbank.GitVersioning/blob/274a42ed3321bcc43632dc970d65677f99a486a9/src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs#L278), this makes sense.

To summarize, my intention is to build from a tagged past commit again in exactly the same way as when the tag was created. GitHub Actions does not allow changing the `GITHUB_REF`, so changing the environment accordingly is not possible.

Options that came to my mind:
* add a `BuildingRef` override by env variable
  * probably change [this line](https://github.com/dotnet/Nerdbank.GitVersioning/blob/274a42ed3321bcc43632dc970d65677f99a486a9/src/NerdBank.GitVersioning/VersionOracle.cs#L63) to something similar to `this.BuildingRef = buildingRefOverrideFromEnv ?? cloudBuild?.BuildingTag ?? cloudBuild?.BuildingBranch ?? context.HeadCanonicalName;`
  * This also adds a lot more flexibility for a lot of other use cases than this specific one.
* add an equivalent of `git tag --points-at HEAD` to [`GitRepository.GetHeadAsReferenceOrSha()`](https://github.com/dotnet/Nerdbank.GitVersioning/blob/274a42ed3321bcc43632dc970d65677f99a486a9/src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs#L278)

Is there any smarter way around this? What do you think?

Complete example workflow runs:
* [Original run triggered by publishing a new github release, works](https://github.com/akademischerverein/ava/actions/runs/3249084677/jobs/5331068018)
* [Re-run on same commit, checked out by tag, does not work (non-public release)](https://github.com/akademischerverein/ava/actions/runs/3695998306/jobs/6259113295)

georg-jung:
Fixes #873.

This implements my second idea described in detail in #873.

My thoughts form [there](https://github.com/dotnet/Nerdbank.GitVersioning/issues/873#issuecomment-1352830730):

> * It would _just work_ without any configuration needed (one might argue it's a breaking change though; could be opt-in to mitigate, wouldn't _just work_ then).
> * It hits the notch that _git is the only source of truth_.
> * With more complicated pipelines and changing environments (e.g. the nbgv step vs. inside a docker build container), it is still straight forward (if `.git` is available, it works) and we don't need to worry about _is this env var for the override available _inside_ the container that performs the build?_
> * It feels natural, because it does what I specified in `version.json` and what I specified there does work in the typical case, where `GITHUB_REF` is not stale.
> * It makes local builds where the tag is checked out equivalent to cloud builds initiated by tag push. Currently `"^refs/tags/v\\d+\\.\\d+"` doesn't have any effect locally if I get it right - it does in CI environments though.

This comes with parsing support for annotated tags for managed git and LibGit2. Lightweight tags pointing HEAD are considered as well as annotated tags. Nested annotated tags that indirectly point at HEAD are not considered (intentionally, changing probably wouldn't be hard). `BuildingRef` value is not changed.

Currently, the output of `nbgv get-version -f json` is extended as follows:

```json
  "VersionHeight": 3,
  "VersionHeightOffset": 0,
  "BuildingRef": "2e6084bfeb07d7f8e426b52aa51bd7517faf5610",
  "BuildingTags": [
    "refs/tags/possiblyASecondTag",
    "refs/tags/v0.7.3",
  ],
  "Version": "0.7.3.11872",
  "CloudBuildAllVarsEnabled": false,
  "CloudBuildAllVars": {
    // ...
    "NBGV_BuildingRef": "2e6084bfeb07d7f8e426b52aa51bd7517faf5610",
    "NBGV_BuildingTags": "System.Collections.Generic.List`1[System.String]",
    "NBGV_Version": "0.7.3.11872",
    // ...
  },
  "CloudBuildVersionVarsEnabled": true,
```

~Obviously the `NBGV_BuildingTags` value isn't here to stay. I thought I'd leave this for discussion, which of these values should be generated and what their value should be.~ I added an `[Ignore]` to skip it in a3089b1e9e3de4f826dfb98bd166a6a31c7fc798.

Probably I made some opinionated decisions when putting this together, don't hesitate to change what you don't like.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/Nerdbank.GitVersioning?pullRequest=876) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/Nerdbank.GitVersioning?pullRequest=876) before we can accept your contribution.<br/><br/>:x: georg-jung [sign now](https://cla.dotnetfoundation.org/dotnet/Nerdbank.GitVersioning?pullRequest=876)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/Nerdbank.GitVersioning?pullRequest=876) it.</sub>

georg-jung:
In [3c93a89](https://github.com/dotnet/Nerdbank.GitVersioning/pull/876/commits/3c93a89145e5f2421e5ba320dfc7a1c44457e650) I added a workaround for `GitPack.TryGetObject` which kind of violates its implicit contract (it throws if it can not get an object of the requested type). I wasn't sure about the best way to fix this. Also this is kind of unrelated to the rest of this PR.

src/NerdBank.GitVersioning/VersionOracle.cs
```diff
@@ -269,10 +277,16 @@ public IEnumerable<string> BuildMetadataWithCommitId
...
-    /// Gets or sets the ref (branch or tag) being built.
+    /// Gets or sets the ref (branch or tag) being built. Just contains a tag if it is known
+    /// that explicitly this tag is build, e.g. in a cloud build context.
     /// </summary>
     public string? BuildingRef { get; protected set; }
 
+    /// <summary>
+    /// Gets or sets a collection of the tags that reference HEAD.
+    /// </summary>
+    public IReadOnlyCollection<string>? BuildingTags { get; protected set; }
```
ronaldbarendse, line 288: To prevent this property from being outputted in the `CloudBuildAllVars`, you can add the `Ignore` attribute:
```suggestion
    [Ignore]
    public IReadOnlyCollection<string>? BuildingTags { get; protected set; }
```

ronaldbarendse: This attribute should also be added to the `VersionOptions` property, since that is now outputted as:
```json
"CloudBuildAllVars": {
  // ...
  "NBGV_VersionOptions": "Nerdbank.GitVersioning.VersionOptions",
  // ...
}
```

ronaldbarendse: To fix the output format, the `CloudBuildAllVars` property body can be updated to support converting different values to strings:

```c#
private const string IsoDateTimeFormat = "yyyy'-'MM'-'dd'T'HH':'mm':'ss.FFFFFFFK";

public IDictionary<string, string> CloudBuildAllVars
{
    get
    {
        var variables = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);

        PropertyInfo[]? properties = this.GetType().GetTypeInfo().GetProperties(BindingFlags.Public | BindingFlags.DeclaredOnly | BindingFlags.Instance);
        foreach (PropertyInfo? property in properties)
        {
            if (property.GetCustomAttribute<IgnoreAttribute>() is null)
            {
                object? propertyValue = property.GetValue(this);
                if (propertyValue is null)
                {
                    continue;
                }

                string value = propertyValue switch
                {
                    IEnumerable<string> values => string.Join(",", values),
                    DateTime dateTime => dateTime.ToString(IsoDateTimeFormat),
                    DateTimeOffset dateTimeOffset => dateTimeOffset.ToString(IsoDateTimeFormat),
                    IFormattable formattable => formattable.ToString(null, CultureInfo.InvariantCulture),
                    _ => propertyValue.ToString() ?? string.Empty,
                };

                variables.Add($"NBGV_{property.Name}", value);
            }
        }

        return variables;
    }
}
```

This also fixes formatting of dates and other values that use the default culture, so instead of outputting:
```json
"CloudBuildAllVars": {
  // ...
  "NBGV_GitCommitDate": "15-12-2022 21:40:21 +00:00",
  // ...
}
```
This will now output it as (similar to the `GitCommitDate` value is the root of the JSON object):
```json
"CloudBuildAllVars": {
  // ...
  "NBGV_GitCommitDate": "2022-12-15T21:40:21+00:00",
  // ...
}
```

ronaldbarendse (changes requested):

src/NerdBank.GitVersioning/VersionOracle.cs
georg-jung: Thanks for pointing me in the `[Ignore]` direction, that seems like the right way to exclude it. On the other hand, excluding might not be what we want as I think `BuildingTags` can provide valuable information to a caller/someone who tries to understand what nbgv does why in a specific case.

`VersionOptions ` is a good point, I agree we should probably handle this in the same way as we handle `BuildingTags` (if we keep both; or: `[Ignore]` `VersionOptions` and handle `BuildingTags` differently).

I also think it's preferrable to use ISO date time formats. However, that seems like a breaking change to me. Also, maybe there is a better way to handle all the different possible property types than to switch on every type that comes to our mind? How to hande enumerable types, such as `BuildingTags`, if we want to include these? Comma-separate and recurse? Maybe json serializing each value on its own is a sane and also very simple and well-defined option?

Happy to implement any of these options - I'd be interested in @AArnott's opinion first though :).

ronaldbarendse: Adding `[Ignore]` will only remove the `NBGV_BuildingTags` item from the `CloudBuildAllVars` JSON array and not set the environment variable when `cloudBuild.setAllVariables` is enabled; the values will still be available in the JSON output as `BuildingTags` (as shown in your example output). The same is true for `VersionOptions`...

Since environment variable values can only be strings and the current implementation simply uses `ToString()`, it uses the current culture, doesn't support formats and can't serialize complex objects (like string arrays/lists).

The current culture issue can be fixed by using `Convert.ToString(value, CultureInfo.InvariantCulture)` instead of `value.ToString()`, but that could potentially also result in a different output format for dates. Since the current output format is unspecified, I don't really see this as being a breaking change though.

Regarding enumerable types, the default environment variable value separator on Windows is a semicolon (``;`), but a colon (`:`) on other OSes. The above code is using a comma (`,`) to separate the tags, but looking at the [git documentation](https://git-scm.com/docs/git-check-ref-format), both semicolons and commas are allowed in tag names, making the colon the only robust separator (for this specific value).

Getting input from @AArnott before implementing any changes would indeed help and avoid doing unnecessary work 👍🏻

georg-jung: I agree that separating the culture concerns from this and creating a second PR makes sense. I also added the suggested `[Ignore]` attribute.

georg-jung:
@AArnott anything I could do from my side to assist/simplify getting this merged?

src/NerdBank.GitVersioning/VersionOracle.cs
AArnott: I agree with *not* exposing the array in an environment variable before someone needs it.

AArnott:
I'm looking at this now and preparing some local changes to push to your PR.

src/NerdBank.GitVersioning/ManagedGit/GitPack.cs
```diff
@@ -152,8 +153,19 @@ public bool TryGetObject(GitObjectId objectId, string objectType, out Stream? va
...
-            value = this.GetObject(offset.Value, objectType);
-            return true;
+            // This try-catch should probably be replaced by a non-throwing GetObject implementation.
+            // This is in turn dependend on a proper GitPackReader.TryGetObject implementation.
+            try
+            {
+                value = this.GetObject(offset.Value, objectType);
+                return true;
+            }
+            catch (GitException gexc)
```
AArnott, line 163: Under what circumstances does this throw where that's unacceptable? Is it for something other than a shallow clone?

src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs
```diff
@@ -648,6 +637,65 @@ public string GetCacheStatistics()
...
+                candidates.Add((tagObjId, refName));
+            }
+        }
+
+        var tags = new List<string>();
+        foreach ((GitObjectId tagObjId, string tagNameCandidate) in candidates)
+        {
+            if (objectId.Equals(tagObjId))
+            {
+                tags.Add(tagNameCandidate);
```
AArnott, line 684: Filtering the tags at this point is quite late. NB.GV functions in some repos with tens of thousands of tags, and the code would have parsed and allocated multiple strings for *all* of them. I think we'll need to optimize this by walking the tags and immediately skipping lines with non-matching object IDs before we allocate anything.

AArnott (changes requested):
Thanks for putting this together. I have several comments I look forward to discussing with you.

src/NerdBank.GitVersioning/ManagedGit/GitAnnotatedTag.cs
```diff
@@ -0,0 +1,73 @@
...
+    public string Tag { get; set; }
+
+    /// <summary>
+    /// Gets or sets the type of the object this tag is pointing to, e.g. "commit" or, for nested tags, "tag".
+    /// </summary>
+    public string Type { get; set; }
+
+    public static bool operator ==(GitAnnotatedTag left, GitAnnotatedTag right)
+    {
+        return Equals(left, right);
```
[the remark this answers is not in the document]
AArnott: I fixed this.

src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs
```diff
@@ -648,6 +637,65 @@ public string GetCacheStatistics()
...
+
+            candidates.Add((tagObjId, lightweightTagName));
+        }
+
+        // packed-refs file
+        foreach (var line in this.EnumeratePackedRefsEntries())
+        {
+            var tagObjId = GitObjectId.Parse(line.Substring(0, 40));
+            var refName = line.Substring(41);
+
```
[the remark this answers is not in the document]
georg-jung: git's source supports that: <https://github.com/git/git/blob/d9d677b2d8cc5f70499db04e633ba7a400f64cbf/refs/packed-backend.c#L542>

> A "record" here is one line for the reference itself and zero or one peel lines that start with '^'.

Thus, if I understand correctly, simply skipping these lines would be functionally equivalent to if they didn't exist, because we'd peel the refs in the next step. Having them does provide us with a way to speed things up though and hits the notch @AArnott mentioned [here](https://github.com/dotnet/Nerdbank.GitVersioning/pull/876#discussion_r1109179767) - we could decide if we're interested in this ref in place. Right?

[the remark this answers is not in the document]
georg-jung: Good to know, thanks! Actually I think this behaviour makes sense and is also in line with what the code in this PR currently does. My take on "tags at HEAD" behaves like `git tag --points-at HEAD` would. Means it includes any lightweight as well as annotated tags pointing at HEAD. But _not_ any tags that point at HEAD transitively. A `git show HEAD` on the other hand does include transitive tags. Not including transitive tags might lead to less lookups => increased performance. In a broken repo, lookup including transitive tags could lead to an infinite loop.

I guess transitive tags are kind of an edge case so my plan was to just not support them at this point. I think specifying "We do what `git tag --points-at HEAD` does" makes sense too. If the need to support transitive tags would arise, a second PR would probably be easier than this first one. What do you think @AArnott?

src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs
georg-jung: I agree that this can be solved without the candidate list and that the number of string allocations can be reduced. The current implementation without tag support [does however allocate multiple strings per packed-ref line too](https://github.com/dotnet/Nerdbank.GitVersioning/blob/5b56c63eb22315152b06bdef353df0ee60116cf3/src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs#L392) and it doesn't seem to be a performance hit or is there some detail I'm missing?

Looking up the target of an annotated tag leads to an IO operation - read the relevant git pack. Thus, I'm wondering if collecting the candidates first wouldn't have a minor impact compared to opening the file.

I don't have any that large repo at hand. If this turns out to have a substantial negative performance impact, it might be an option to make annotated tag support opt-in/out, as reading the packed-refs file itself (plus the files in `refs/tags`) is sufficient for lightweight tag lookup.

I'm working on a commit to remove the candidate list and reduce the string allocations though.

src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs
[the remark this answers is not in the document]
georg-jung: Makes sense. You might want to take a look at my two recent commits in that regard.

src/NerdBank.GitVersioning/ManagedGit/GitPack.cs
georg-jung: `GitPackReader.GetObject` [will throw](https://github.com/dotnet/Nerdbank.GitVersioning/blob/5b56c63eb22315152b06bdef353df0ee60116cf3/src/NerdBank.GitVersioning/ManagedGit/GitPackReader.cs#L71) if we ask for the wrong object type. We can not _always ask for the right object type_ though, because we don't know if a tag points e.g. at a commit or at another tag. Thus, this is a rather common path. If we have a tag and a GitObjectId that does not equal the one we are looking for, the tag could either be unrelated and point any other commit or it could be an annotated tag that in turn does point to the GitIbjectId we are looking for. I thought about refactoring this to "proper" Try*-pattern functions that will throw under no circumstances but I found this to be a rather big and non-straightforward change.

AArnott:
The build failure reproduces in main, so it's unlikely to be due to this PR. I don't know how that happened, but I'll be investigating it.

georg-jung:
@AArnott I noticed that my implementation led to a rather substantial decrease in performance - the more refs exist for a repo, the bigger the decrease. Thus, I did some further research on that topic and was able to optimize my tag approach so that there isn't a performance hit anymore in most cases. git does actually guarantee the existance of peel-lines in the `packed-refs` file if the header of the file specifies so. This feature exists at least [since 2017](https://github.com/git/git/blob/72d4a9a721d519982cbc97c36f3fcf33b4103ed4/refs/packed-backend.c#L169) so it isn't anything new - probably longer, I just didn't look further.

Wrapping up, with my new optimizations:

There isn't a performance hit if:
* ManagedGit is used
* The `packed-refs` file supports peel lines, which it does for some years
* I'd expect the chance for a modern `packed-refs` file is especially high for repos with a large number of refs, where the performance decrease in nbgv would be relevant, given this is a performance optimization within git itself for probably the exact same performance reasons we're seeing here

There still is a performance hit with my implementation if:
* the LibGit2 backend is used
* the `packed-refs` file supports neither the `peeled` nor the `fully-peeled` trait

I guess this eases the situation quite much. If you see a need for further optimizations let me know.

AArnott:
Thank you, @georg-jung. I'm working on fixing the latent build breaks in main and v3.5 that happened due to changes in the Azure Pipelines agents, then I can come back to your PR. 
In the meantime, thanks for the perf fixes. I wonder if we can negate _all_ negative perf impact for most folks by only searching tags when the `publicReleaseRefSpec` actually includes tags. If it doesn't, there's no reason to compute that data, right?

georg-jung:
I thought about that idea too. But as `publicReleaseRefSpec` just contains regexes in general, they could match against tags without explicitly containing "tag", couldn't they?

AArnott:
Yes, I suppose they could. Although I wouldn't feel _too_ bad about only supporting tags if `tags/` appeared in the regex expression. 
But libgit2 is meant as an escape hatch and isn't a mainline scenario anymore. I'll take a look at the code soon and see if I have any strong feelings one way or the other after that.

AArnott:
@georg-jung I was prepared to merge this, but happened to hit an error related to the change. Can you check [this out](https://dev.azure.com/andrewarnott/OSS/_build/results?buildId=7930&view=logs&j=22fea640-1099-5f32-ec5d-316ad83f359a&t=e5b82db5-077e-546d-323a-6c8f7f824b90&l=28)?

```
[20:42:24] Error: Unhandled exception: Newtonsoft.Json.JsonSerializationException: Error getting value from 'Tags' on 'Nerdbank.GitVersioning.VersionOracle'.
 ---> System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Nerdbank.GitVersioning.ManagedGit.GitObjectId.ParseHex(ReadOnlySpan`1 value) in /home/vsts/work/1/s/src/NerdBank.GitVersioning/ManagedGit/GitObjectId.cs:line 131
   at Nerdbank.GitVersioning.ManagedGit.GitAnnotatedTagReader.ReadObject(ReadOnlySpan`1 line) in /home/vsts/work/1/s/src/NerdBank.GitVersioning/ManagedGit/GitAnnotatedTagReader.cs:line 100
   at Nerdbank.GitVersioning.ManagedGit.GitAnnotatedTagReader.Read(ReadOnlySpan`1 tag, GitObjectId sha) in /home/vsts/work/1/s/src/NerdBank.GitVersioning/ManagedGit/GitAnnotatedTagReader.cs:line 72
   at Nerdbank.GitVersioning.ManagedGit.GitAnnotatedTagReader.Read(Stream stream, GitObjectId sha) in /home/vsts/work/1/s/src/NerdBank.GitVersioning/ManagedGit/GitAnnotatedTagReader.cs:line 48
   at Nerdbank.GitVersioning.ManagedGit.GitRepository.LookupTags(GitObjectId objectId) in /home/vsts/work/1/s/src/NerdBank.GitVersioning/ManagedGit/GitRepository.cs:line 678
   at Nerdbank.GitVersioning.Managed.ManagedGitContext.get_HeadTags() in /home/vsts/work/1/s/src/NerdBank.GitVersioning/Managed/ManagedGitContext.cs:line 64
   at Newtonsoft.Json.Serialization.ExpressionValueProvider.GetValue(Object target)
```

georg-jung:
@dotnet-policy-service agree

georg-jung:
Hey @AArnot,

sorry for the delay. As a first measure, I tried to reproduce this and created a branch in my fork that incorporates the recent fixes in main (to make the pipelines in my repo work again) as well as this PR's code.

I [wasn't able](https://dev.azure.com/georg-jung/Nerdbank.GitVersioning/_build/results?buildId=1842&view=results) to reproduce though. The stacktrace points to a piece of code that wasn't touched by this PR - altough I don't really understand the line number in the stacktrace, as that line only reads `var objectId = default(GitObjectId);`. So I decided to ignore the line number for now and looked at that method in whole but I wasn't able to find any error and still came to the conclusion that this piece wasn't touched here.

What are the chances this was just some temporary hick up with the pipeline? An IO error in the checkout?

Does anything stop us form just running that pipeline again and see if its reproducible?

AArnott:
>  The stacktrace points to a piece of code that wasn't touched by this PR - altough I don't really understand the line number in the stacktrace, as that line only reads var objectId = default(GitObjectId);

Are you looking at the same error? [This is what I see](https://dev.azure.com/andrewarnott/OSS/_build/results?buildId=7930&view=logs&j=0bc77094-9fcd-5c38-f6e4-27d2ae131589&t=d0f091e0-7dba-547a-a9e4-e48ea7b20a3f&l=28):

```
[20:46:40] Error: Unhandled exception: Newtonsoft.Json.JsonSerializationException: Error getting value from 'Tags' on 'Nerdbank.GitVersioning.VersionOracle'.
 ---> System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Nerdbank.GitVersioning.ManagedGit.GitObjectId.ParseHex(ReadOnlySpan`1 value) in D:\a\1\s\src\NerdBank.GitVersioning\ManagedGit\GitObjectId.cs:line 131
   at Nerdbank.GitVersioning.ManagedGit.GitAnnotatedTagReader.ReadObject(ReadOnlySpan`1 line) in D:\a\1\s\src\NerdBank.GitVersioning\ManagedGit\GitAnnotatedTagReader.cs:line 100
   at Nerdbank.GitVersioning.ManagedGit.GitAnnotatedTagReader.Read(ReadOnlySpan`1 tag, GitObjectId sha) in D:\a\1\s\src\NerdBank.GitVersioning\ManagedGit\GitAnnotatedTagReader.cs:line 72
   at Nerdbank.GitVersioning.ManagedGit.GitAnnotatedTagReader.Read(Stream stream, GitObjectId sha) in D:\a\1\s\src\NerdBank.GitVersioning\ManagedGit\GitAnnotatedTagReader.cs:line 48
   at Nerdbank.GitVersioning.ManagedGit.GitRepository.LookupTags(GitObjectId objectId) in D:\a\1\s\src\NerdBank.GitVersioning\ManagedGit\GitRepository.cs:line 678
   at Nerdbank.GitVersioning.Managed.ManagedGitContext.get_HeadTags() in D:\a\1\s\src\NerdBank.GitVersioning\Managed\ManagedGitContext.cs:line 64
   at Newtonsoft.Json.Serialization.ExpressionValueProvider.GetValue(Object target)
```

That looks extremely relevant to this PR, considering this PR introduces `GitAnnotatedTagReader` which makes up several frames in the callstack.

> What are the chances this was just some temporary hick up with the pipeline? An IO error in the checkout?

Checkout is long behind this failure. And exactly the same error happened on both Windows and linux, so no, I don't think we can right this one off.

Sometimes a failure can occur because of the particular way a git database is represented on disk, so that can be a challenge to repro on another clone of the repo. But it's not something we want to sweep under the rug, because if it can happen in a PR build, it can certainly happen on a dev box among the many thousands that this code runs on.

AArnott:
I've been looking this over to try to figure out what may be going wrong, and it occurs to me that we don't have a single new test for the tags functionality in this PR. Sorry I missed this before, but we will need some tag tests to accept this PR as well.

AArnott:
I added a test for Tags to your PR. If you can think of anything else to test, you're welcome to. We have the facility to check in whole repo databases in order to be able to test particular git db formats (pack files, peeled lines, etc) too.

AArnott:
... and the re-build failed at exactly the same place, in the same way. So, at least it's reproducible. I can't reproduce it locally (yet).

AArnott:
I pushed a commit that will capture the entire clone, including .git directory, next time it fails. That should help us repro it.

AArnott:
Yay! I have a local repro by downloading the drop I changed the pipeline to produce on failure. I'm investigating now.

src/NerdBank.GitVersioning/ManagedGit/GitAnnotatedTagReader.cs
```diff
@@ -0,0 +1,133 @@
...
+
+    private static GitObjectId ReadObject(ReadOnlySpan<byte> line)
+    {
+        // Format: object d8329fc1cc938780ffdd9f94e0d364e0ea74f579\n
+        // 48 bytes:
+        //  object: 6 bytes
+        //  space: 1 byte
+        //  hash: 40 bytes
+        //  \n: 1 byte
+        Debug.Assert(line.Slice(0, ObjectStart.Length).SequenceEqual(ObjectStart));
```
AArnott, line 97: In the repro of the failure in Azure Pipelines, this is the first sign of failure. The assertion fails in debug builds.

AArnott: The actual exception is thrown from `GitObjectId.ParseHex` on the second `ReverseHexDigits` reference because `value[i] - '0']` evaluates to -38, which is of course an invalid array index.

AArnott (changes requested):
This is a legitimate failure, introduced by the `GitAnnotatedTagReader` code, apparently.

You can repro by following these steps:

1. Download the `drop` artifact from [here](https://dev.azure.com/andrewarnott/OSS/_build/results?buildId=7951&view=artifacts&pathAsName=false&type=publishedArtifacts) and use 7-zip to extract it. Complete all the rest of these steps within the unzipped directory.
2. Open VS to the solution within the unzipped directory.
1. Add `Debugger.Launch()` to the `GitObjectId.ParseHex` method.
1. Run `dotnet publish -o ../nerdbank-gitversioning.npm/out/nbgv.cli/tools/net6.0/any` from the `src\nbgv` directory.
3. Run `yarn build` from the `src\nerdbank-gitversioning.npm` directory.
1. When the JIT debugger attach dialog appears, have the VS with your unzipped solution open attach to the debuggee.
2. Configure the debugger to break on first chance exceptions, at least for `IndexOutOfRangeException`.
3. `GitException` is thrown a lot due to the issue of relying on exceptions to recognize the wrong object type. You can turn off break on first chance exceptions for this particular exception type.
4. Hit F5 till you encounter a debug assertion failure and/or the `IndexOutOfRangeException` that causes the ultimate failure.

Can you please look into this?

Also, given the `GitException` is thrown about 200 times before this failure, this is likely going to need to be addressed. I already have a local code change that I hope will address this, although it causes other failures, so I'll work through that after we get this fatal error resolved.

src/NerdBank.GitVersioning/ManagedGit/GitAnnotatedTagReader.cs
georg-jung: Thanks for the great repro instructions! I'm looking into this now. It seems like something is fed to `GitAnnotatedTagReader` that is not actually an annotated tag. That leads to the failure because the object is then obviously not in the format the `GitAnnotatedTagReader` expects to parse.

georg-jung: I think fixing this is loosely related to 

> Also, given the GitException is thrown about 200 times before this failure, this is likely going to need to be addressed. I already have a local code change that I hope will address this, although it causes other failures, so I'll work through that after we get this fatal error resolved.

My first idea is that either the value or key of `GitPackCache` needs to contain the object type. Currently the type is just [determined](https://github.com/dotnet/Nerdbank.GitVersioning/blob/f66943e12e66454d3e3ac705a4f8e650e4531ff3/src/NerdBank.GitVersioning/ManagedGit/GitPackReader.cs#L43) when the object is actually read, but not for cache access situations.

I think this might be a larger change though and it is not closely related to this PR's topic. I'd be willing to further contribute (I'm on vacation 2023-03-25 till 2023-04-02) in this direction but probably it's best to discuss how to proceed first.

AArnott: Yes, I see what you mean. Nice find. Let's chat when you get back from vacation. If I have a chance in the meantime, maybe I'll be able to fix it.

georg-jung: I'm back now 👍🏻

AArnott: Great. I haven't had time to look at this yet. Do you have time to look into the fix? As you say it's unrelated (but important to) this PR, if you have a fix, consider sending it as a separate PR which we could merge first, then re-validate this PR on a fresher merge.

src/NerdBank.GitVersioning/ManagedGit/GitPack.cs
AArnott: It appears this is no longer thrown a lot. :)

AArnott (approved):

PR #905, opened 2023-03-10
.NET documentation starter example include

sengiv:
include an simple example to help beginners access `ThisAssembly` during runtime


sengiv:
@dotnet-policy-service agree

AArnott:
Thanks for contributing. But while this makes the doc 50% longer, I don't see it as adding much or any value. We already include the class declaration itself. Why do folks need to know how to access a const field on a class? And why do we need to demonstrate doing it for every single one of them?

sengiv:
good point, no worries, will close it, thanks

PR #911, opened 2023-03-20
Add support for `NBGV_ThisAssemblyNamespace` property.

### LINKED ISSUE #555: Allow changing the name or namespace of ThisAssembly class
We encountered the following scenario where `ThisAssembly caused a build break.
1. Project A has `ThisAssembly` generated and also `[InternalsVisibleTo]` for project B
2. Project B also has `ThisAssembly`. When it tries to use it, build fails because of ambiguous ThisAssembly class.

The solution to this issue may be either naming ThisAssembly class differently, or placing it inside a namespace (root namespace of the project, perhaps). For backward compatibility, this should be opt-in.

alexrp:
Closes #555


alexrp:
@AArnott How do you feel about the approach here?

Also, how would you suggest I test this? I'm not really familiar with NBGV's test harness.

AArnott:
@KalleOlaviNiemitalo MSBuild does not distinguish between null or an empty string as property values, nor an unset value vs. a value set to an empty string. An MSBuild task property that is set to an msbuild property will never be null.

AArnott:
The PR looks ineffective for the other languages as well, since only F# emits a namespace. The other languages just no-op for the emit namespace instruction.

AArnott (changes requested):
@alexrp you can add the tests to [this file](https://github.com/dotnet/Nerdbank.GitVersioning/blob/main/test/Nerdbank.GitVersioning.Tests/AssemblyInfoTest.cs). Please test all the languages. There is already a test for each language, so leave those alone I suppose but duplicate them with your optional property set.
Please make sure the task works properly given task inputs that are set to the empty string (indicating the msbuild property is not set).

alexrp:
> The PR looks ineffective for the other languages as well, since only F# emits a namespace. The other languages just no-op for the emit namespace instruction.

Not sure I follow here. There's code to emit namespaces for all of them. It's just that the F# case is handled specially in `EmitNamespaceIfRequired()`.

src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs
```diff
@@ -119,7 +121,8 @@ public string BuildCode()
                 this.generator.AddBlankLine();
                 this.generator.AddAnalysisSuppressions();
                 this.generator.AddBlankLine();
-                this.generator.EmitNamespaceIfRequired(this.RootNamespace ?? "AssemblyInfo");
+                this.generator.EmitNamespaceIfRequired(this.ThisAssemblyNamespace ?? this.RootNamespace ?? "AssemblyInfo");
```
alexrp, line 124: So is the conclusion that this should be changed to use `string.IsNullOrEmpty()` for both of these properties?

AArnott: Ya, that would probably be appropriate.

AArnott:
> There's code to emit namespaces for all of them. 

No, I don't think so. The tests themselves prove that no namespace is emitted for C# or VB, but there is for F#. Issue #555 was originally filed because this class is always in the same global namespace for all projects.

alexrp:
> The tests themselves prove that no namespace is emitted for C# or VB

But that's expected, right? The PR makes it so that, for C#/VB, a namespace is only emitted if the project explicitly sets `NBGV_ThisAssemblyNamespace` to some non-empty value. For F#, a namespace is indeed always emitted as was the case prior to the PR.

I *could* make it always emit a namespace, but I figured that might be a bad idea for compatibility reasons.

alexrp:
> If I'm wrong, can you point me to the code that emits the namespace?

Here (and similarly for VB):

https://github.com/dotnet/Nerdbank.GitVersioning/pull/911/files#diff-784868d214d0f8f1823e743bdc0b7b56b4e899e08201e712ccb2bc9e35594977L756-R766

I put the code in those methods out of convenience since for C#/VB it requires opening *and* closing a namespace, unlike F#. (Also, I figured we shouldn't rely on file-scoped namespaces since people might be using older C# versions?) I could add a new pair of `StartThisAssemblyNamespace`/`EndThisAssemblyNamespace` methods if you think that'd be clearer.

alexrp:
OK, here's take 2, now with tests.

test/Nerdbank.GitVersioning.Tests/AssemblyInfoTest.cs
```diff
@@ -90,6 +90,67 @@ namespace AssemblyInfo
         Assert.Equal(expected, built);
     }
 
+    [SkippableTheory(typeof(PlatformNotSupportedException))]
+    [InlineData(null, "MyRootNamespace")]
+    [InlineData("", "MyRootNamespace")]
+    [InlineData("MyCustomNamespace", null)]
+    [InlineData("MyCustomNamespace", "")]
+    [InlineData("MyCustomNamespace", "MyRootNamespace")]
```
AArnott, line 98: Can you add one more case here, where *both* inputs are null/empty so that we hit the `AssemblyInfo` namespace case?

AArnott (changes requested):
Looks much better. Thank you. Just one small change please.

test/Nerdbank.GitVersioning.Tests/AssemblyInfoTest.cs
alexrp: The test above this one is testing a few different things, but I believe that specific case is also covered by it?

AArnott (approved):

AArnott:
Thanks again.

PR #918, opened 2023-04-07
Compatibility improvement for WiX Toolset V4 projects.

### LINKED ISSUE #913: Nerdbank.GitVersioning used with WixToolset Bundle projects
I have the following error message when i want to use Nerdbank.GitVersioning on WixToolset Bundle (output .exe) project:
 _CodeDomProvider not available for language: wix. No version info will be embedded into assembly._

The property <NBGV_EmitThisAssemblyClass>false</NBGV_EmitThisAssemblyClass> it doesn't solve the problem.

Do you think this can be handled in [AssemblyVersionInfo.cs](https://github.com/dotnet/Nerdbank.GitVersioning/blob/6d30af47025b6af6e094e6054b67dc12a18b054f/src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs)?

maniglia:
In WiX Toolset projects the language is set to _wix_, in these cases the _GenerateAssemblyVersionInfo_ is set to false.

Fixes #913


AArnott (approved):
Thanks!

PR #996, opened 2023-11-09
Allow customizing commit message pattern in `prepare-release` command

### LINKED ISSUE #991: New feature - Command "prepare-release"  with a paremeter for add a prefix to the commit's
**Please Describe The Problem To Be Solved:**
In our solution, for every commit generated, we have a format. In this case, we use the "JIRA" card format, for example "LM-xxx-{commit message}". So when the command "prepare-release" is executed, it generates two commits. We need to add this prefix to every commit. So we ask if we can collaborate to add this new feature, which should be an optional parameter to the command.

Actual value : "Set version to '2.1'"
Expected value: "LM-1450 - Set version '2.1'"

fmacavilca:
By default, the `prepare-release` command generates a commit message with the format "Set version to {version}". So with this new option, You can add a prefix or suffix to the default commit message (which will replace just the version).

For example, running the following command on `master`

```
nbgv prepare-release --commit-message-pattern "Custom commit message pattern - {0} custom message"
```

So your commit message is going to be this:

```
Custom commit message pattern '1.0' custom message
```
**Note:** Must include {0} in the pattern to represent the version.

closes #991

### DIFF
--- a/doc/nbgv-cli.md
+++ b/doc/nbgv-cli.md
@@ -167,7 +167,7 @@ In that case, the `NewBranch` property will be `null`.
 
 ### Customizing the `prepare-release` commit message
 
-By default, the `prepare-release` command generates commit message with this format "Set version to {version}". So with this new option. You can add a prefix or suffix to the default commit message
+By default, the `prepare-release` command generates a commit message with the format "Set version to {version}". So with this new option, You can add a prefix or suffix to the default commit message (which will replace just the version).
 
 For example, running the following command on `master`
 
@@ -178,9 +178,9 @@ nbgv prepare-release --commit-message-pattern "Custom commit message pattern - {
 So your commit message is going to be this:
 
 ```
-Custom commit message pattern - Set version to '1.0' custom message
+Custom commit message pattern '1.0' custom message
 ```
-**Note:** You must consider that commit message pattern must contain only one {0} and valid commit message characters.
+**Note:** Must include {0} in the pattern to represent the version.
 
 ## Creating a version tag
 
--- a/doc/nbgv-cli.md
+++ b/doc/nbgv-cli.md
@@ -167,9 +167,10 @@ In that case, the `NewBranch` property will be `null`.
 
 ### Customizing the `prepare-release` commit message
 
-By default, the `prepare-release` command generates a commit message with the format "Set version to {version}". So with this new option, You can add a prefix or suffix to the default commit message (which will replace just the version).
+By default, the `prepare-release` command generates a commit message with the format "Set version to {version}".
+A switch allows you to customize the commit message, using `{0}` as a placeholder for the version.
 
-For example, running the following command on `master`
+For example, running the following command:
 
 ```
 nbgv prepare-release --commit-message-pattern "Custom commit message pattern - {0} custom message"
@@ -178,9 +179,8 @@ nbgv prepare-release --commit-message-pattern "Custom commit message pattern - {
 So your commit message is going to be this:
 
 ```
-Custom commit message pattern '1.0' custom message
+Custom commit message pattern - 1.0 custom message
 ```
-**Note:** Must include {0} in the pattern to represent the version.
 
 ## Creating a version tag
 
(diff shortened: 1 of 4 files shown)


fmacavilca:
@dotnet-policy-service agree

FernandoCalmet (approved):

wchoque (approved):

doc/nbgv-cli.md
```diff
@@ -165,6 +165,23 @@ For each branch, the following properties are provided:
 **Note:** When the current branch is already the release branch for the current version, no new branch will be created.
 In that case, the `NewBranch` property will be `null`.
 
+### Customizing the `prepare-release` commit message
```
AArnott, line 168: Thank you for the documentation update!
It makes me wonder... maybe when the message is customized, the `{0}` should be replaced with only the version itself, allowing total customization of the message.

src/nbgv/Program.cs
```diff
@@ -217,6 +219,7 @@ private static Parser BuildCommandLine()
                 var nextVersion = new Option<string>("--nextVersion", "The version to set for the current branch. If omitted, the next version is determined automatically by incrementing the current version.");
                 var versionIncrement = new Option<string>("--versionIncrement", "Overrides the 'versionIncrement' setting set in version.json for determining the next version of the current branch.");
                 var format = new Option<string>(new[] { "--format", "-f" }, $"The format to write information about the release. Allowed values are: {string.Join(", ", SupportedFormats)}. The default is {DefaultOutputFormat}.").FromAmong(SupportedFormats);
+                var commitMessagePattern = new Option<string>("--commit-message-pattern", "Custom pattern to add a prefix or suffix to the default commit message.");
```
AArnott, line 222: Can you add to the description something to indicate that `{0}` is expected to appear somewhere in the argument?

src/nbgv/Program.cs
```diff
@@ -763,11 +767,24 @@ private static Task<int> OnPrepareReleaseCommand(string project, string nextVers
                 return Task.FromResult((int)ExitCodes.UnsupportedFormat);
             }
 
+            // validate commit message pattern
+            if (AreCurlyBracesBalanced(commitMessagePattern))
```
AArnott, line 771: Let's not get into the business of manual validation of the message. Let's leave it to `string.Format` and (if we can) catch any exception thrown in order to print the error message and return the error code.

AArnott (changes requested):
I like it. Just a few changes please. The design change idea (about reducing what `{0}` is replaced with) is just an idea. It's at your discretion.

AArnott (approved):
Thanks for this!

FernandoCalmet (approved):

src/nbgv/Program.cs
```diff
@@ -69,6 +70,7 @@ private enum ExitCodes
             ShallowClone,
             InternalError,
             InvalidTagNameSetting,
+            InvalidunformattedCommitMessage,
```
choquewilber, line 73: @AArnott I think it should be Invalid`U`nformattedCommitMessage

wchoque, line 73: @AArnott I think it should be Invalid`U`nformattedCommitMessage to follow PascalCase convention

AArnott: Whoops. That was a search and replace result. Yes, I'll fix.

fmacavilca:
@AArnott, do you know when it will be published in the new release?

AArnott:
I can issue a prerelease to nuget.org if that will suffice for you.

fmacavilca:
Yes, please. It will be great. Thanks

AArnott:
https://www.nuget.org/packages/Nerdbank.GitVersioning/3.7.48-alpha

doc/nbgv-cli.md
ronaldbarendse: Shouldn't the placeholder be named `{version}` instead, both to be more descriptive and align with the release tag and branch name formats?

The current code also uses `string.Format()` without any validation or exception handling, so any unescaped `{...}` values that are not `{0}` will cause a `FormatException` 😢

PR #1005, opened 2023-11-15
Make `$(NBGV_ThisAssemblyNamespace)` default to `$(RootNamespace)` as per the docs

inthemedium:
Currently, the default for `$(NBGV_ThisAssemblyNamespace)` is empty.

[The docs indicate it should be `$(RootNamespace)`](https://github.com/dotnet/Nerdbank.GitVersioning/blob/0e809cfdd62297a622fed18215b2c82960f27733/doc/msbuild.md?plain=1#L49).

### DIFF
--- a/test/Nerdbank.GitVersioning.Tests/SomeGitBuildIntegrationTests.cs
+++ b/test/Nerdbank.GitVersioning.Tests/SomeGitBuildIntegrationTests.cs
@@ -627,7 +627,7 @@ public abstract class SomeGitBuildIntegrationTests : BuildIntegrationTests
         AssemblyProductAttribute assemblyProduct = assembly.GetCustomAttribute<AssemblyProductAttribute>();
         AssemblyCompanyAttribute assemblyCompany = assembly.GetCustomAttribute<AssemblyCompanyAttribute>();
         AssemblyCopyrightAttribute assemblyCopyright = assembly.GetCustomAttribute<AssemblyCopyrightAttribute>();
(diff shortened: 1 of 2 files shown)


test/Nerdbank.GitVersioning.Tests/SomeGitBuildIntegrationTests.cs
```diff
@@ -627,7 +627,7 @@ public async Task AssemblyInfo(bool isVB, bool includeNonVersionAttributes, bool
         AssemblyProductAttribute assemblyProduct = assembly.GetCustomAttribute<AssemblyProductAttribute>();
         AssemblyCompanyAttribute assemblyCompany = assembly.GetCustomAttribute<AssemblyCompanyAttribute>();
         AssemblyCopyrightAttribute assemblyCopyright = assembly.GetCustomAttribute<AssemblyCopyrightAttribute>();
-        Type thisAssemblyClass = assembly.GetType("ThisAssembly") ?? assembly.GetType("TestNamespace.ThisAssembly");
+        Type thisAssemblyClass = assembly.GetType("ThisAssembly") ?? assembly.GetType("TestNamespace.TestNamespace.ThisAssembly");
```
inthemedium, line 630: Should I update something else?

AArnott:
Thanks for calling out the mismatching code and docs. In this case, the docs are wrong. We cannot change the default namespace as that would be a substantially breaking change for existing users.

PR #1012, opened 2023-12-31
AssemblyVersionInfo: Use `global::` for all references to types in `System.*` namespaces.

alexrp:
This is necessary to avoid lookup errors when a user's `NBGV_ThisAssemblyNamespace` ends with `.System`, and probably other similar situations.


AArnott (approved):
Looks good. Thanks!

PR #1024, opened 2024-01-16
Change F# AssemblyInfo generation to always include a do() after the …

### LINKED ISSUE #1022: Compile errors in an F# project when setting NBGV_EmitThisAssemblyClass to false
Hi,
I've been using GitVersioning in some F# projects, and it's all been working.

However, I tried setting the ```NBGV_EmitThisAssemblyClass``` property to false, and then started getting a build error from the generated Version.fs file:
```
error FS0010: Incomplete structured construct at or before this point in definition
```

Looking at the code, the version file with the default options is
```fsharp
namespace xxx
[<assembly: System.Reflection.AssemblyVersionAttribute("0.7.0.0")>]
[<assembly: System.Reflection.AssemblyFileVersionAttribute("0.7.59.18254")>]
[<assembly: System.Reflection.AssemblyInformationalVersionAttribute("0.7.59-prerelease+474e680183")>]
do()
#if NETSTANDARD || NETFRAMEWORK || NETCOREAPP
[<System.CodeDom.Compiler.GeneratedCode("Nerdbank.GitVersioning.Tasks","3.6.133.12845")>]
#endif
#if NET40_OR_GREATER || NETCOREAPP2_0_OR_GREATER || NETSTANDARD2_0_OR_GREATER
[<System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage>]
#endif
type internal ThisAssembly() =
   ...
```
But with NBGV_EmitThisAssemblyClass = false it's
```fsharp
namespace xxx
[<assembly: System.Reflection.AssemblyVersionAttribute("0.7.0.0")>]
[<assembly: System.Reflection.AssemblyFileVersionAttribute("0.7.59.18254")>]
[<assembly: System.Reflection.AssemblyInformationalVersionAttribute("0.7.59-prerelease+474e680183")>]
```
without the trailing ```do()```, which seems to be the problem.

Just from a quick look at the GitVersioning code, it looks like the ```do()``` gets added by the ```StartThisAssemblyClass``` at https://github.com/dotnet/Nerdbank.GitVersioning/blob/2b5ba90c74dc5d3e9300e8079d6ed511536000a9/src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs#L741C21-L741C21 and then that isn't called when ThisAssembly is diabled?

This is with version 3.6.133, though the 3.7.48-alpha seems the same.

Numpsy:
…version attributes

Fixes #1022 

As it stands, assembly version information for F# when ThisAssembly generation is disabled generates things like
image
because the ```do()``` after the attributes is required and isn't added.

There is a unit test for this case, but I don't think it's quite right as it expects the ```do()``` to be a header for the ```#if NETSTANDARD || NETFRAMEWORK || NETCOREAPP``` bit instead of a footer for the assembly version attributes.

### DIFF
--- a/src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs
+++ b/src/Nerdbank.GitVersioning.Tasks/AssemblyVersionInfo.cs
@@ -397,6 +397,8 @@ namespace Nerdbank.GitVersioning.Tasks
                     this.generator.DeclareAttribute(typeof(AssemblyCopyrightAttribute), this.AssemblyCopyright);
                 }
             }
+
+            this.generator.EndAssemblyAttributes();
         }
 
         private List<KeyValuePair<string, (object Value, bool EmitIfEmpty /* Only applies to string values */)>> GetFieldsForThisAssembly()
@@ -658,6 +660,10 @@ namespace Nerdbank.GitVersioning.Tasks
             {
             }
 
+            internal virtual void EndAssemblyAttributes()
+            {
+            }
+
             internal abstract void DeclareAttribute(Type type, string arg);
 
             internal abstract void StartThisAssemblyClass();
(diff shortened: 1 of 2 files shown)


AArnott (approved):
Thanks!

PR #1043, opened 2024-04-09
Move ProjectReference into a conditional import

### LINKED ISSUE #1040: Warning about PrivateP2PCaching.proj
I have the same issue as #787. If I understand correctly, this issue should have been circumvented for v3.6 by #847. However, I still get the issue with v3.6.133. I'm not using a graph build, AFAIK:

`Directory.Build.props`:
```xml
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="Current" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <ItemGroup>
    <PackageReference Include="Nerdbank.GitVersioning" Condition="!Exists('packages.config')">
      <PrivateAssets>all</PrivateAssets>
      <Version>3.6.133</Version>
    </PackageReference>
  </ItemGroup>
</Project>
```

AArnott:
This makes the `ProjectReference` item itself completely undiscoverable for project systems that enumerate references using the MSBuild API that gets items regardless of condition, which appears to be how the legacy project system (csproj.dll) does it at least sometimes.

Fixes #1040

### DIFF
--- a/src/Nerdbank.GitVersioning.Tasks/build/MSBuildTargetCaching.targets
+++ b/src/Nerdbank.GitVersioning.Tasks/build/MSBuildTargetCaching.targets
@@ -33,14 +33,14 @@
       <Visible>false</Visible>
       <PrivateAssets>all</PrivateAssets>
     </NBGV_CachingProjectReference>
-
-    <!-- Declare a P2P so that "msbuild -graph -isolate" doesn't complain when we use the MSBuild task to invoke our inner shared project.
-         This causes a lot of problems (https://github.com/dotnet/Nerdbank.GitVersioning/issues?q=label%3Amsbuild-p2p+) with projects
-         that expect to understand all their own ProjectReferences though, so only define it when the user is running a graph build. -->
-    <ProjectReference Include="@(NBGV_CachingProjectReference)" Condition="'$(IsGraphBuild)'=='true'">
-      <NBGV_InnerProject>true</NBGV_InnerProject>
-    </ProjectReference>
   </ItemGroup>
+
+  <!-- Declare a P2P so that "msbuild -graph -isolate" doesn't complain when we use the MSBuild task to invoke our inner shared project.
+       This causes a lot of problems (https://github.com/dotnet/Nerdbank.GitVersioning/issues?q=label%3Amsbuild-p2p+) with projects
+       that expect to understand all their own ProjectReferences though, so only define it when the user is running a graph build.
+       Rather than condition the item (which the legacy project system 'sees through' using MSBuild APIs),
+       we condition an import so that the item will be utterly undiscoverable by the project system. -->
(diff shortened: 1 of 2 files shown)


AArnott:
@thomas-ravkilde  Can you test this? You can download the nupkg from the PR build artifacts. Let me know if you need more instructions.

thomas-ravkilde:
@AArnott Sorry for the delay. I'm afraid I need some more instructions. I thought I could just include prereleases in the VS NuGet Package Manager but that doesn't appear to be the case. I see v3.6.79-alpha as the latest prerelease on v3.6. Where do I find the PR build artifacts?

AArnott:
Perfectly understandable, and thank you for trying. The instructions for consuming PR build artifacts is [near the bottom of our README](https://github.com/dotnet/Nerdbank.GitVersioning?tab=readme-ov-file#how-do-i-consume-the-latest-changes-prior-to-their-release-on-nugetorg).
But even that only carries CI builds. For a PR build, you'd need to download the nupkg artifact from the PR build and then point your nuget.config file to the directory that contains the nupkg. You can get it from the `deployables-Windows` artifact [here](https://dev.azure.com/andrewarnott/OSS/_build/results?buildId=9332&view=artifacts&pathAsName=false&type=publishedArtifacts).

thomas-ravkilde:
Ah, I see. Thanks, I couldn't figure out how to get to the [PR build artifacts](https://dev.azure.com/andrewarnott/OSS/_build/results?buildId=9332&view=artifacts&pathAsName=false&type=publishedArtifacts).

Unfortunately, the change doesn't fix the issue:
image

AArnott:
Thanks for testing and reporting back.
I have no idea how it would be failing at this point unless there's something very peculiar in your setup.
Can you send me a minimal repro?

thomas-ravkilde:
Hm... Very interesting.
- I created a new .NET Framework 4.8 Console project. --> No issue.
- I removed all projects in my solution to create a minimal repro. --> Still got the warning.
- Cleaned the solution and rebuilt. --> Still got the warning.
- Ran `git clean -xdf`, cleaned the solution and rebuilt. --> Now there's no warning.
So it appears that your change did actually fix the issue but that there was something sticky in the build process.

AArnott:
Progress. Great.

Can you verify with one more step: roll back to the latest release of NB.GV and verify that even after `git clean`  the warning comes back?

AArnott:
@thomas-ravkilde I'm holding off merging this awaiting your confirmation on my last request.

thomas-ravkilde:
@AArnott OK. I finally got around to testing this thoroughly.
- With v3.6.133 the error disappears after `git clean -xdf`.
- It appears to arise from installation via the nbgv .NET client tool, but not via NuGet.

So I guess you shouldn't merge. I'm sorry to have wasted your time.

AArnott:
Great. I'm glad to hear the original fix for this was adequate. I don't know why `nbgv install` would have caused the problem to initially appear, however.

PR #1082, opened 2024-08-26
Update NuGet.PackageManagement to v6.3.4

### LINKED ISSUE #1081: nuget.packaging 6.3.1 is being flagged by vulnerability scanners
I'm building Docker containers that include nbgv 3.6 and the container images are being flagged by JFrog XRay with https://github.com/NuGet/NuGet.Client/security/advisories/GHSA-68w7-72jg-6qpp due to the presence of nuget.packaging v6.3.1

Apparently this vulnerabiliy is fixed in versions 5.11.6, 6.0.6, 6.3.4, 6.4.3, 6.6.2, 6.7.1, 6.8.1

If we could upgrade NuGet.PackageManagement to 6.3.4 I think that would mitigate the issue (as I believe that's what is causing NuGet.Packaging to be included)

flcdrg:
Fixes #1081


AArnott (approved):
Thanks

PR #1090, opened 2024-09-03
Ensure we use System.Drawing.Common 4.7.2

### LINKED ISSUE #1089: 3.6 is using vulnerable version of System.Drawing.Common
According to JFrog XRay, the current 3.6 release is using System.Drawing.Common v4.7.0, which has https://nvd.nist.gov/vuln/detail/CVE-2021-24112

I don't see this as a direct dependency, so assume it must be being pulled in as a transient one.

Upgrading to 4.7.2 or 5.0.3 should apparently resolve this.

flcdrg:
Fixes #1089

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -29,6 +29,7 @@
     <PackageVersion Include="NuGet.PackageManagement" Version="6.3.4" />
     <PackageVersion Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
     <PackageVersion Include="System.Diagnostics.Tools" Version="4.3.0" />
+    <PackageVersion Include="System.Drawing.Common" Version="4.7.2" />
     <PackageVersion Include="System.Memory" Version="4.5.5" />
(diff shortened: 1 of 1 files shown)


src/NerdBank.GitVersioning/Nerdbank.GitVersioning.csproj
```diff
@@ -16,6 +16,7 @@
     <PackageReference Include="Newtonsoft.Json" />
     <PackageReference Include="Validation" />
     <PackageReference Include="Nerdbank.GitVersioning.LKG" />
+    <PackageReference Include="System.Drawing.Common" />
```
AArnott, line 19: We don't need this because we set `CentralPackageTransitivePinningEnabled` to `true`. Can you drop this change?

flcdrg: Done

AArnott (approved):

PR #1100, opened 2024-10-03
ReleaseManager: do not reset VersionHeightOffset if it equals to -1

### LINKED ISSUE #1058: Issue with nbgv prepare-release Command Removing versionHeightOffset in ReleaseFlow Branching Strategy
Description:
I'm trying to apply the ReleaseFlow branching strategy to our project and using NerdBank.GitVersioning as our versioning tool. However, I'm encountering an issue where the versionHeightOffset property is removed from the version.json file after running the nbgv prepare-release command.

Current Configuration:
Here's the version.json file I'm using:

```json
{
  "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/main/src/NerdBank.GitVersioning/version.schema.json",
  "version": "1.2-SNAPSHOT.{height}",
  "versionHeightOffset": -1,
  "publicReleaseRefSpec": [
    "^refs/heads/master$",
    "^refs/heads/release/v\\d+(?:\\.\\d+)+$"
  ],
  "release": {
    "branchName": "release/v{version}",
    "firstUnstableTag": "SNAPSHOT"
  }
}
```
Problem:
After running the nbgv prepare-release command, the versionHeightOffset property is removed, resulting in the following version.json:

```json
{
  "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/main/src/NerdBank.GitVersioning/version.schema.json",
  "version": "1.3-SNAPSHOT.{height}",
  "publicReleaseRefSpec": [
    "^refs/heads/master$",
    "^refs/heads/release/v\\d+(?:\\.\\d+)+$"
  ],
  "release": {
    "branchName": "release/v{version}",
    "firstUnstableTag": "SNAPSHOT"
  }
}
```
Objective:
I want the version to be 1.2.0 after running the prepare-release command, instead of 1.2.1. I understand that the patch version 1 is based on the height, but I prefer it to be 0 after the release preparation. Also, I want the patch number to increase with each new commit to the release branch.

What I Tried:
I know I can manually set the version property like "version": "1.3.0-SNAPSHOT.{height}", but this isn't ideal as I want the patch number to automatically increase with new commits to the release branch.

Question:
How can I achieve the desired versioning behavior where the versionHeightOffset is maintained, and the version becomes 1.2.0 after running the prepare-release command?

DmitryZhelnin:
Fixes #1058 

Add additional check if `VersionHeightOffset` equals to `-1` during execution of `prepare-release` command

### DIFF
--- a/src/NerdBank.GitVersioning/ReleaseManager.cs
+++ b/src/NerdBank.GitVersioning/ReleaseManager.cs
@@ -280,7 +280,7 @@ public class ReleaseManager
 
         if (!EqualityComparer<SemanticVersion>.Default.Equals(versionOptions.Version, newVersion))
         {
-            if (versionOptions.VersionHeightPosition.HasValue && SemanticVersion.WillVersionChangeResetVersionHeight(versionOptions.Version, newVersion, versionOptions.VersionHeightPosition.Value))
+            if (versionOptions.VersionHeightOffset != -1 && versionOptions.VersionHeightPosition.HasValue && SemanticVersion.WillVersionChangeResetVersionHeight(versionOptions.Version, newVersion, versionOptions.VersionHeightPosition.Value))
             {
                 // The version will be reset by this change, so remove the version height offset property.
                 versionOptions.VersionHeightOffset = null;
--- a/test/Nerdbank.GitVersioning.Tests/ReleaseManagerTests.cs
+++ b/test/Nerdbank.GitVersioning.Tests/ReleaseManagerTests.cs
@@ -638,6 +638,44 @@ public abstract class ReleaseManagerTests : RepoTestBase
         Assert.Equal(expectedReleaseVersionOptions, releaseVersion);
     }
 
+    [Fact]
+    public void PrepareRelease_DoesNotResetNegativeVersionHeightOffset()
+    {
+        // create and configure repository
+        this.InitializeSourceControl();
+
+        var initialVersionOptions = new VersionOptions
+        {
+            Version = SemanticVersion.Parse("1.0-beta"),
+            VersionHeightOffset = -1,
+        };
+
+        var expectedReleaseVersionOptions = new VersionOptions
+        {
+            Version = SemanticVersion.Parse("1.0"),
+            VersionHeightOffset = -1,
+        };
+
+        var expectedMainVersionOptions = new VersionOptions
+        {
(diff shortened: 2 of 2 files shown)


src/NerdBank.GitVersioning/ReleaseManager.cs
```diff
@@ -280,7 +280,7 @@ private void UpdateVersion(LibGit2Context context, SemanticVersion oldVersion, S
 
         if (!EqualityComparer<SemanticVersion>.Default.Equals(versionOptions.Version, newVersion))
         {
-            if (versionOptions.VersionHeightPosition.HasValue && SemanticVersion.WillVersionChangeResetVersionHeight(versionOptions.Version, newVersion, versionOptions.VersionHeightPosition.Value))
+            if (versionOptions.VersionHeightOffset >= 0 && versionOptions.VersionHeightPosition.HasValue && SemanticVersion.WillVersionChangeResetVersionHeight(versionOptions.Version, newVersion, versionOptions.VersionHeightPosition.Value))
```
AArnott, line 283: I know you're doing exactly what I asked for. In retrospect though, I wonder if we're over-shooting here. If someone had `-200` here, that would almost certainly be for a specific reason that no longer applies after a version bump. And it would be invalid after too, because -199 isn't a valid version number.
Let's change this to only skip on *exactly* `-1`. This fits your scenario, your test, and is the only valid negative number (since it will result in an initial version of 0).
```suggestion
            if (versionOptions.VersionHeightOffset != -1 && versionOptions.VersionHeightPosition.HasValue && SemanticVersion.WillVersionChangeResetVersionHeight(versionOptions.Version, newVersion, versionOptions.VersionHeightPosition.Value))
```

AArnott (changes requested):
Thanks for this. One change please.

src/NerdBank.GitVersioning/ReleaseManager.cs
DmitryZhelnin: Thanks for your reply! Yes, `-1` value looks to be the only valid case.

AArnott: Fair point, @KalleOlaviNiemitalo. We can adjust this further if we get feedback from users that use it. We might require a CLI switch to prevent resetting the value.

AArnott (approved):

PR #1103, opened 2024-10-09
fix GitCommitDate being author date rather than commit date

### LINKED ISSUE #827: GitCommitDate is author date rather than commit date
The value of the MSBuild property `$(GitCommitDateTicks)` comes from the Git **author** date, even though the name and API documentation indicate that it should come from the Git **commit** date. The commit date would be more useful in our workflow in which we often cherry-pick commits that are several months old. If the commit date is old, then it is easy to know for sure that the deployed version cannot include any changes that were implemented after that date. The author date cannot be used for this purpose.

Reproduced with Nerdbank.GitVersioning 3.4.240, but the code looks like it works like that in v3.5.113 too:

<https://github.com/dotnet/Nerdbank.GitVersioning/blob/a908c6d9d2ed9b032d63ff885b4ccc97fc32046d/src/NerdBank.GitVersioning/LibGit2/LibGit2Context.cs#L50>
<https://github.com/dotnet/Nerdbank.GitVersioning/blob/a908c6d9d2ed9b032d63ff885b4ccc97fc32046d/src/NerdBank.GitVersioning/Managed/ManagedGitContext.cs#L53>
<https://github.com/dotnet/Nerdbank.GitVersioning/blob/a908c6d9d2ed9b032d63ff885b4ccc97fc32046d/src/NerdBank.GitVersioning.Tests/GitContextTests.cs#L43>

DmitryZhelnin:
Fill GitCommitDate with actual value of commit date.
Add GitCommitAuthorDate property for convenience.

fixes #827

### DIFF
--- a/src/NerdBank.GitVersioning/DisabledGit/DisabledGitContext.cs
+++ b/src/NerdBank.GitVersioning/DisabledGit/DisabledGitContext.cs
@@ -24,6 +24,8 @@ internal class DisabledGitContext : GitContext
 
     public override DateTimeOffset? GitCommitDate => null;
 
+    public override DateTimeOffset? GitCommitAuthorDate => null;
+
     public override string? HeadCanonicalName => null;
 
     public override IReadOnlyCollection<string>? HeadTags => null;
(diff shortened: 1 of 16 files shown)


AArnott (approved):
Thanks so much!

AArnott:
Actually something about this caused a bunch of tests to fail. @DmitryZhelnin can you look into that?

DmitryZhelnin:
> Actually something about this caused a bunch of tests to fail. @DmitryZhelnin can you look into that?

Sure, I'll investigate.

DmitryZhelnin:
@AArnott I'm sorry, there were some missing changes. Tests should be OK now.

PR #1120, opened 2024-11-26
Add dotnet sdk configuration for dependabot

JamieMagee:
This configuration will allow dependabot to create patch PRs for the .NET SDK. See https://github.blog/changelog/2024-11-19-dependabot-can-now-perform-version-updates-for-the-net-sdk/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,11 @@ updates:
   directory: /src/nerdbank-gitversioning.npm
   schedule:
     interval: monthly
+- package-ecosystem: dotnet-sdk
+  directory: /
+  schedule:
+    interval: monthly
+  ignore:
+    - dependency-name: "*"
+      update-types:
+        - version-update:semver-major


.github/dependabot.yml
```diff
@@ -11,3 +11,11 @@ updates:
...
   schedule:
     interval: monthly
+- package-ecosystem: dotnet-sdk
+  directory: /
+  schedule:
+    interval: monthly
+  ignore:
+    - dependency-name: "*"
+      update-types:
+        - version-update:semver-major
```
AArnott, line 21: I'm not familiar with the behavior of this last bit. I haven't been able to find much documentation on the `dotnet-sdk` "package-ecosystem" and its options.

JamieMagee: The last section would prevent major version SDK upgrades automatically. Check out this blog post for more information: https://devblogs.microsoft.com/dotnet/using-dependabot-to-manage-dotnet-sdk-updates/

AArnott: Cool. Thanks for the link.
However in the case of this repo specifically, there are very specific SDK version requirements and unfortunately the SDK version must match in 3 different places in the repo, so a dependabot-created PR to bump the SDK will _always_ break the repo (even for servicing fixes).
That, and I tend to upgrade the SDK version with a merge from Library.Template, which not only updates the global.json file but also the `.devcontainer\Dockerfile` file to match (something I wish dependabot would do). 

So I'm going to thank you for contributing, and close this particular PR without merging.

Issue #1173, opened 2025-02-14
Latest F# compiler warns on autogenerated Version.fs


Smaug123:
https://github.com/dotnet/fsharp/pull/17871 made it a compile warning to specify an invalid #nowarn suppression in an F# file. The default Version.fs generated file contains a nowarn, CA2243 (sorry, I'm on my phone, will provide more details and possibly a fix when I get to a real computer).

I believe the correct fix is simply not to nowarn in the F# generated file.

Numpsy:
Same here fwiw:
```
PackagingTools\artifacts\obj\PackageViewer\release_win-x64\PackageViewer.Version.fs(11,9): warning FS0203: Invalid warning number 'CA2243'
```

CA2243 looks like a Roslyn analyzer, so I don't think it would run on F# code anyway

Smaug123:
Thanks for the fix! @AArnott What's the release process for this? I can of course work around the bug, but last night Dependabot's automatic PRs became blocked in most of my repos when they attempted to upgrade the .NET SDK. Would it be easy to cut a release of Nerdbank.GitVersioning?

AArnott:
@Smaug123 You can always consume our CI builds as [described here](https://dotnet.github.io/Nerdbank.GitVersioning/docs/faq.html#how-do-i-consume-the-latest-changes-prior-to-their-release-on-nugetorg).

If you need the package on nuget.org, I could ship a 3.8 pre-release to nuget.org. Or if you need a _stable_ versioned release we could backport the fix to 3.7.

Smaug123:
If you could release the fix in some stable form, that would be great - it's hit pretty much every dotnet project I'm involved with at this point.

AArnott:
Does the [v3.8.38-alpha release](https://github.com/dotnet/Nerdbank.GitVersioning/releases/tag/v3.8.38-alpha) not cut it?

Numpsy:
fwiw I've updated a few projects to the 3.8 alpha and that's fixed the warnings for me

Smaug123:
Sorry for the noise - the alpha release does work for me, thanks.

PR #1174, opened 2025-02-14
Don't try to disable CA2243 warnings in the generated version info files for F#

Numpsy:
refs #1173

Maybe it'd be easier to just make AddAnalysisSuppressions() a no-op for F# for now, but I thought maybe overriding WarningCodesToSuppress to an empty collection and then handling that case would be more flexible in case it actually does need to supress any other warnings later.


Smaug123 (approved):
Looks locally plausible, though I know nothing about the implementation of NBGV!

AArnott (approved):
Thanks

PR #1190, opened 2025-03-23
Add option to set / skip CloudBuildNumber

### LINKED ISSUE #1189: Option to not emit vso[build.updatebuildnumber] to prevent pipeline warnings
My organization disallows setting the build number via the logging command:

```
##[warning]##vso[build.updatebuildnumber] is not allowed in this step due to policy restrictions. Please reference documentation (http://go.microsoft.com/fwlink/?LinkId=817296)
```

However, running `nbgv cloud` is still useful because it's allowed to set the environment variables, which I can use to set the build number in a supported manner.

It might be niche, but is it possible to add an option to prevent emitting this log line, since it results in a warning that I don't believe there's any other way to suppress?

Or if you have any thoughts or other workarounds, please let me know. Thanks!

MattKotsenas:
Fixes #1189 

This PR adds the ability to set / skip the CloudBuildNumber in both the Cake task and the CLI.

## Common
- Adds a new bool `cloudBuildNumber` to `CloudCommand::SetBuildVariables`
 
## Cake.GitVersioning

- Adds a new `CloudBuildNumber` to `GetVersioningCloudSettings` with the default value of `true`
- Makes the default values of `AllVariables` and `CommonVariables` explicit (but no change)

## nbgv CLI

- Adds a new command line switch `--skip-cloud-build-number` / `-b` which if specified sets `cloudBuildNumber` to `false`.

## Tests

- There weren't any automated tests to cover this scenario, so added some basic tests

### DIFF
--- a/test/Nerdbank.GitVersioning.Tests/CommandTests.cs
+++ b/test/Nerdbank.GitVersioning.Tests/CommandTests.cs
@@ -12,10 +12,12 @@ public class CommandTests : RepoTestBase
     {
     }
 
-    // TODO: This is tightly coupled to the cloud build service. Can I use a FakeCloudBuildService instead?
     [Theory, CombinatorialData]
-    public void CloudCommand_CloudBuildNumber([CombinatorialValues("VisualStudioTeamServices", "TeamCity", "Jenkins")] string ciSystem, bool setCloudBuildNumber)
+    public void CloudCommand_CloudBuildNumber(bool setCloudBuildNumber)
     {
+        const string ciSystem = "VisualStudioTeamServices";
+        const string buildNumberSyntax = "##vso[build.updatebuildnumber]";
+
         var outWriter = new StringWriter();
         var errWriter = new StringWriter();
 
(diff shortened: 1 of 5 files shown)


MattKotsenas:
I felt bad not writing tests for this, but I'm not happy with the tests that I wrote:

- They're tied to individual cloud providers, which feels wrong
- They test by side effect, assuming that if all + common + cloudBuildNumber are all off, the output will always be empty

test/Nerdbank.GitVersioning.Tests/CommandTests.cs
```diff
@@ -0,0 +1,43 @@
...
+using Xunit;
+
+public class CommandTests : RepoTestBase
+{
+    public CommandTests(ITestOutputHelper logger)
+        : base(logger)
+    {
+    }
+
+    // TODO: This is tightly coupled to the cloud build service. Can I use a FakeCloudBuildService instead?
```
AArnott, line 15: Since the logic you're testing here doesn't vary by CI platform, I'd suggest you just hard-code to one. Then look specifically for the logging command you know to expect from that one.
If you keep the positive and negative test as you have it here, if the CI platform ever changes their logging command and NB.GV updates to compensate, your test here will fail because the positive test won't see the expected logging command, and we can update the test at the same time as we update the library.

A FakeCloudBuildService would be fine, but I don't know of a way to do that from a test project, and I don't want to add it to the shipping library.

AArnott (changes requested):
Thanks for this. I left a comment on a suggested test change, but otherwise I'm happy to take this.

I also removed the ` = false;` property initializers you added since I don't care for explicitly setting defaults.

test/Nerdbank.GitVersioning.Tests/CommandTests.cs
MattKotsenas: Updated. If this isn't what you had in mind, please let me know. Thanks!

AArnott (approved):

PR #1253, opened 2025-08-17
When generating the GitCommitDate field in the AssemblyInfo for F#, d…

Numpsy:
…on't add an explicit 'new'

Just a thought after running [FSharpLint](https://github.com/fsprojects/FSharpLint) on some projects using GitVersioning and getting
```
Usage of `new` keyword here is redundant.
Error on line 31 starting at column 41
  static member internal GitCommitDate = new global.System.DateTime(638905509020000000L, global.System.DateTimeKind.Utc)
```
because the ```new``` is optional is F# and isn't usually used for things like this
image

This is rather trivial and you could say that FSharpLint shouldn't care about ```GeneratedCode```, but I'm putting this here if anyone is interested.


AArnott:
@Numpsy did you leave this in draft mode because you intend to add more changes?

PR #1256, opened 2025-08-28
Only do Android version check on applications

dotMorten:
If your version for a class library exceeds 124.255.65535, the android target will fail.
This limitation of versioning should only apply to applications (Maui apps will have `OutputType` = `Exe`), and not class libraries.

Added an additional condition to the error message to only fail on applications, and unblocks building other project types.


AArnott (approved):
Thanks

PR #1264, opened 2025-09-08
Fix `nbgv set-version` to write to the best version.json file in scope

### LINKED ISSUE #1257: `dotnet nbgv set-version` modifies root `version.json` when executed in inherited `version.json` context
## Problem
In a parent-child `version.json` structure, when running `nbgv set-version X`, it always updates the parent `version.json` file even when executed in the context of the child `version.json` one.

This not only causes values to be updated in the wrong place (messing the root file), but it also causes future commands to crash because it incorrectly sets `inherited: true` on the base file, causing the following exception on subsequent commands (such as `nbgv get-version`):
```
dotnet nbgv get-version
Unhandled exception: System.InvalidOperationException: "C:\git\affinitiv\dotnet-extensions\version.json" inherits from a parent directory version.json file but none exists.
   at Nerdbank.GitVersioning.VersionFile.GetWorkingCopyVersion(String startingDirectory, String& actualDirectory) in D:\a\1\s\src\NerdBank.GitVersioning\VersionFile.cs:line 228
```

## Context
I have a requirement to manually set the version of my project in Github actions. This project is one of many in a monorepo where there is a root `version.json` file with basic settings such as `publicReleaseRefSpec` and `cloudBuild` settings, and separate inherited `version.json` files per "component" folder, where I store the version of each component in the repo.

My structure is like this:
```
├── version.json
├── src/
     ├── Component1/
     │      └── version.json
     └── Component2/
            └── version.json
      ...
```

The issue occurs no matter how I provide the component folder context:
- If I navigate to `src/Component1`, then run `nbgv set-version X`
- If `nbgv set-version X --project src/Component1` directly in the root

## Technical Details
I grabbed the repo to debug the issue to understand what was going on.

Because the config exists, it always lands on the first `if` block here and persists the settings to `actualDirectory`, which always seems to return the base file path no matter how the action is called:

https://github.com/dotnet/Nerdbank.GitVersioning/blob/543debd6b79be79e02c080bc77a17326fe744231/src/nbgv/Program.cs#L695-L715

Image

## Workaround
The only thing that comes to mind right now is to just avoid using inherited `version.json` files and replicate the common configuration across all component-specific `version.json`, which is obviously less than ideal considering the main purpose of the shared root file today is to avoid the duplication of settings.

AArnott:
This required reworking how version.json files are discovered to report back on where certain version.json files were. In particular, we had these requirements that were not previously being met:

* We have to know where the nearest version.json file is _that specifies a `version` property_, since that's the file that `nbgv set-version` will be writing to.
* We have to get a deserialized `VersionOptions` object that has _not_ been merged with its parent when `inherit: true` is set. This is so that when we rewrite the file, we only serialize the same properties that came from that file rather than 'flattening' the inheritance hierarchy by including properties brought in from the parent.

Rather than special case these particular requirements, I updated the version.json search algorithm to allow the caller to specify different combinations of requirements and get locations of potentially multiple files.

Fixes #1257

### DIFF
--- a/src/NerdBank.GitVersioning/DisabledGit/DisabledGitVersionFile.cs
+++ b/src/NerdBank.GitVersioning/DisabledGit/DisabledGitVersionFile.cs
@@ -14,9 +14,9 @@ internal class DisabledGitVersionFile : VersionFile
 
     protected new DisabledGitContext Context => (DisabledGitContext)base.Context;
 
-    protected override VersionOptions? GetVersionCore(out string? actualDirectory)
+    protected override VersionOptions? GetVersionCore(VersionFileRequirements requirements, out VersionFileLocations locations)
     {
-        actualDirectory = null;
+        locations = default;
         return null;
     }
 }
--- a/src/NerdBank.GitVersioning/LibGit2/LibGit2GitExtensions.cs
+++ b/src/NerdBank.GitVersioning/LibGit2/LibGit2GitExtensions.cs
@@ -594,7 +594,7 @@ public static class LibGit2GitExtensions
             {
                 try
                 {
-                    options = ((LibGit2VersionFile)this.context.VersionFile).GetVersion(commit, this.context.RepoRelativeProjectDirectory, this.blobVersionCache, out string? actualDirectory);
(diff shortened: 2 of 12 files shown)


AArnott:
@KalleOlaviNiemitalo I'd love to get your review on this too.

PR #1271, opened 2025-09-18
Use and preserve server.json wherever it is found

AArnott:


AArnott:
> Okay, a .NET SDK target was preventing the logic from working correctly and it is required that .NET 10 SDK is used for the new PackageType. After resolving that, the flow works

@timheuer Which target was it? Is it something NB.GV needs to accommodate or was it unique to your situation?

PR #1413, opened 2026-07-09
Only bump version height for filtered-path changes against the first parent

olegtk:
### Problem

When `pathFilters` are configured, the version-height walk marks a commit as height-relevant if the filtered path differs from **any** of its parents:

```csharp
commit.Parents.Any(parent => ContainsRelevantChanges(Compare(parent.Tree, commit.Tree, ...)))
```

For a non-fast-forward merge of a **stale branch** whose copy of the filtered path lagged the mainline, the merge tree is identical to its **first parent** (the mainline) but still differs from the **second parent** (the stale branch). The any-parent check therefore counts the merge as relevant and bumps the height, even though the merge introduced no change to the filtered path.

The practical consequence: routine PR merges of stale branches inflate the version height of a path-filtered project, so its version drifts upward with no actual change under the filtered path.

### Fix

Compare the filtered path against the **first parent only** — the mainline the commit extends. Changes contributed by other parents were already counted in those parents' own heights, which flow into the result via `maxHeightAmongParents`, so they are not lost. A root (no-parent) commit is still relevant if it introduces anything under the filtered path.

Applied to both backends:
- `src/NerdBank.GitVersioning/LibGit2/LibGit2GitExtensions.cs`
- `src/NerdBank.GitVersioning/Managed/GitExtensions.cs`

### Test

Added `GetVersionHeight_MergeOfStaleBranchThatDidNotTouchFilteredPathDoesNotBumpHeight` (runs against both backends): branches before the filtered folder advances, edits the folder on the mainline, edits only a file **outside** the folder on the stale branch, then non-fast-forward merges the stale branch back. The merge leaves the filtered folder identical to its first parent, so the height must stay at 2. This test fails before the fix (height 3) and passes after.

### DIFF
--- a/src/NerdBank.GitVersioning/LibGit2/LibGit2GitExtensions.cs
+++ b/src/NerdBank.GitVersioning/LibGit2/LibGit2GitExtensions.cs
@@ -501,13 +501,17 @@ public static class LibGit2GitExtensions
                         ? null
                         : includePaths;
 
-                // If the diff between this commit and any of its parents
-                // touches a path that we care about, bump the height.
+                // Bump the height only if this commit changed a filtered path relative to its
+                // first parent (the mainline this commit extends). Changes contributed by other
+                // parents were already counted in those parents' heights, which flow in via
+                // maxHeightAmongParents; comparing against every parent instead would count merge
+                // commits that never touched the filtered path (e.g. merges of a stale branch
+                // whose filtered path lagged the mainline), inflating the version height.
                 // A no-parent commit is relevant if it introduces anything in the filtered path.
                 bool relevantCommit =
                     commit.Parents.Any()
-                        ? commit.Parents.Any(parent => ContainsRelevantChanges(commit.GetRepository().Diff
-                            .Compare<TreeChanges>(parent.Tree, commit.Tree, diffInclude, DiffOptions)))
+                        ? ContainsRelevantChanges(commit.GetRepository().Diff
+                            .Compare<TreeChanges>(commit.Parents.First().Tree, commit.Tree, diffInclude, DiffOptions))
                         : ContainsRelevantChanges(commit.GetRepository().Diff
                             .Compare<TreeChanges>(null, commit.Tree, diffInclude, DiffOptions));
 
--- a/src/NerdBank.GitVersioning/Managed/GitExtensions.cs
+++ b/src/NerdBank.GitVersioning/Managed/GitExtensions.cs
@@ -149,22 +149,19 @@ internal static class GitExtensions
 
             if (pathFilters is not null)
             {
-                // If the diff between this commit and any of its parents
-                // touches a path that we care about, bump the height.
-                bool relevantCommit = false, anyParents = false;
-                foreach (GitObjectId parentId in commit.Parents)
+                // Bump the height only if this commit changed a filtered path relative to its
+                // first parent (the mainline this commit extends). Changes contributed by other
+                // parents were already counted in those parents' heights, which flow in via
+                // maxHeightAmongParents; comparing against every parent instead would count merge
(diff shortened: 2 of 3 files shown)


olegtk:
@AArnott — I'd appreciate your guidance on intended semantics before I invest in a real fix here.

**Observed behavior.** With `pathFilters` set, `GetCommitHeight` marks a commit relevant when its filtered tree differs from **any** parent. A non-fast-forward merge of a *stale* branch (branched before the filtered path advanced on the mainline, and never touched that path) yields a merge whose filtered tree equals its **first** parent but differs from the **second** (older) parent. So the merge bumps the path-filtered height even though it changed nothing under the filtered path. In practice, routine PR merges of stale branches drift a path-filtered project's version upward.

**The catch.** The obvious "compare against first parent only" fix (what's on this branch) breaks `GetVersion_PathFilterPlusMerge`, which is structurally identical (merge equals first parent, differs from second parent) but *expects the bump*. The difference between the two is ancestry, not tree state:

- `GetVersion_PathFilterPlusMerge`: the second parent's filtered content **diverged** from the merge base (`bar` vs the base) and the merge overrode it → should count.
- stale-branch: the second parent's filtered content **equals** the merge base (it only lagged) → should not count.

So a correct rule seems to be: *a merge is relevant to the filtered path iff its filtered tree differs from the first parent, or a non-first parent's filtered tree differs from the merge base of the parents.* That needs merge-base computation per merge — available via `FindMergeBase` in the LibGit2 backend, but the **Managed backend has no merge-base support**, so this would be a sizable cross-backend change.

**Questions:**
1. Is the current any-parent merge relevance intentional (as `GetVersion_PathFilterPlusMerge` implies), i.e. is the stale-branch bump working as designed?
2. If you'd consider excluding merely-lagging parents, is the merge-base-aware rule above the semantics you'd want — and would you accept adding merge-base computation to the Managed backend (with an eye on the height-walk perf tests), or is there a lighter approach you'd prefer?

Happy to implement whichever direction you point me toward. This PR is a draft purely to make the tension concrete; the code on the branch is illustrative and not proposed as-is.

AArnott:
> Is the current any-parent merge relevance intentional 

I'm inclined to say _yes_. To git, the parent ordering isn't significant. It only appears significant to us because there's a convention that the first parent is the target being merged into, and git hosting platforms that offer pull requests always use the target branch as the first parent, conventionally. But some of these systems allow fast-forward merges, and some allow direct pushes. These not-so-uncommon variants can _flip_ the parent order such that the target branch ends up as the *second* parent.

So as a matter of predictability and accuracy, NB.GV has never considered parent order as part of version height counting. And outside of the pathFilters scenario, this works out fine. It just means that a non-fast-forward merge of a PR adds one more commit and therefore increments the version height by one.
But when pathFilters gets involved (an idea that I've resisted from before we even added the feature), things get far more complicated and people get more particular about not wanting unexpected version height increments. Your scenario seems _incredibly_ common though, which makes me want to think more deeply about it. It also makes me wonder how people who like path filters haven't run into this and noticed before.

AArnott:
> If you'd consider excluding merely-lagging parents,

Possibly. Especially if we can do it without a bias for the _first_ parent.

Another concern is if any policy change results in lower version height calculations (as this would), we must find a way to avoid decrementing the version height for old commits. If you had version 1.2.300 today and then upgraded NB.GV to get this feature, you mustn't get 1.2.295 tomorrow. The only time we've made a policy change in how we compute version height such that it might compute a smaller number, we made a major version increment of NB.GV and prominently documented it so that people could make one-time adjustments to `versionHeightOffset` in their version.json file to avoid the downgrade. How many people ended up doing that? I don't know.
Or we could have an opt-in for the new policy via a new property in version.json. That would allow us to avoid the smaller version height since old commits would not have opted in.

PR #1436, opened 2026-08-07
Emit placeholder Git metadata when Git engine is disabled

### LINKED ISSUE #1192: NBGV_GitEngine=Disabled does not generate all properties of ThisAssembly type
I was happy to see that with `NBGV_GitEngine=Disabled` there is a way to disable git operations. Sadly not all properties of `ThisAssembly` are then generated and my build breaks.

I use this environment variable to make it possible to use Dependabot. There you always get shallow clones, which breaks Nerdbank.GitVersioning.
An alternate solution for that specific problem would be super welcome.

The behavior that `ThisAssembly` is incomplete when disabled seems like a bug.

## Reproduction
1. Use `ThisAssembly.GitCommitId` in code.
2. Set `NBGV_GitEngine=Disabled`
3. `dotnet build`

AArnott:
Disabling the Git engine currently omits Git-backed `ThisAssembly` members, which can break compilation for projects that reference them. Keep the generated API shape stable by emitting explicit placeholder metadata whenever Git access is disabled.

The string metadata reports `Unavailable from a shallow git clone`, while the machine-readable commit dates use the recognizable Back to the Future timestamp of October 26, 1985 at 1:21 AM UTC. Builds emit suppressible warning `NBGV1001` so consumers cannot silently mistake these placeholders for real source metadata; users can opt out through `MSBuildWarningsAsMessages`.

Integration coverage verifies the generated members, values, warning code, and warning suppression.

Fixes #1192

### DIFF
--- a/src/Nerdbank.GitVersioning.Tasks/GetBuildVersion.cs
+++ b/src/Nerdbank.GitVersioning.Tasks/GetBuildVersion.cs
@@ -11,6 +11,12 @@ namespace Nerdbank.GitVersioning.Tasks
 {
     public class GetBuildVersion : ContextAwareTask
     {
+        private const string UnavailableGitValue = "Unavailable from a shallow git clone";
+        private const string UnavailableGitWarningCode = "NBGV1001";
+
+        // October 26, 1985 at 1:21 AM UTC, when the DeLorean first traveled through time.
+        private const string UnavailableGitDateTicks = "626347344600000000";
+
         /// <summary>
         /// Initializes a new instance of the <see cref="GetBuildVersion"/> class.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/Nerdbank.GitVersioning.Tasks/GetBuildVersion.cs
```diff
@@ -286,6 +292,24 @@ protected override bool ExecuteInner()
...
+                    this.Log.LogWarning(
+                        subcategory: null,
+                        warningCode: UnavailableGitWarningCode,
+                        helpKeyword: null,
+                        file: null,
+                        lineNumber: 0,
+                        columnNumber: 0,
+                        endLineNumber: 0,
+                        endColumnNumber: 0,
+                        message: "Git information is unavailable because the git engine is disabled. ThisAssembly.GitCommitId, ThisAssembly.GitCommitDate, and ThisAssembly.GitCommitAuthorDate contain placeholder values.");
```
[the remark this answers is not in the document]
AArnott: Updated the warning to describe Git-related MSBuild properties and ThisAssembly members collectively, so it covers GitCommitIdShort without maintaining a fragile member list.


> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Nerdbank.GitVersioning.Tasks/GetBuildVersion.cs
```diff
@@ -11,6 +11,12 @@ namespace Nerdbank.GitVersioning.Tasks
 {
     public class GetBuildVersion : ContextAwareTask
     {
+        private const string UnavailableGitValue = "Unavailable from a shallow git clone";
```
KalleOlaviNiemitalo, line 14: I'd prefer plain "Unavailable" here because I disable NBGV in the Debug configuration to speed up incremental builds in IDE even though the Git clone is not shallow.

src/Nerdbank.GitVersioning.Tasks/GetBuildVersion.cs
```diff
@@ -11,6 +11,12 @@ namespace Nerdbank.GitVersioning.Tasks
...
+        private const string UnavailableGitWarningCode = "NBGV1001";
+
+        // October 26, 1985 at 1:21 AM UTC, when the DeLorean first traveled through time.
+        private const string UnavailableGitDateTicks = "626347344600000000";
```
KalleOlaviNiemitalo, line 18: The [Mon Sep 17 00:00:00 2001](https://github.com/git/git/blob/745601a9a94110d74769ab605ccd4f61339758d2/Documentation/git-format-patch.adoc#:~:text=Mon%20Sep%2017%2000:00:00%202001) timestamp from `git format-patch` would be more recognisable to users of Git.

AArnott: Can you tell me more about that? Why is that date significant to git users?

KalleOlaviNiemitalo: It's hardcoded in the From_ line of each `git format-patch` output file. <https://github.com/git/git/blob/745601a9a94110d74769ab605ccd4f61339758d2/log-tree.c#L508>

However, that date being recognisable does not necessarily mean it is appropriate to use here.

Re the DeLorean timestamp, I'm not sure whether the film specified a time zone, but I doubt it would have been UTC.

AArnott: That doc you linked to suggests that that particular timestamp is "to help programs like file(1) to recognize that the file is an output from this command", so reusing that timestamp for other purposes would erode the purpose that that timestamp was originally chosen.

As for the time zone, you're right. It would have been Pacific Daylight Time (UTC-7). So I'll fix that. :)

AArnott:
@KalleOlaviNiemitalo see #1465 for fixes

PR #1512, opened 2026-09-09
Re-enable code coverage from linux runners

fhnaseer:


Directory.Packages.props
```diff
@@ -43,7 +43,9 @@
     <PackageVersion Include="Xunit.Combinatorial" Version="2.1.41" />
   </ItemGroup>
   <ItemGroup Label="Library.Template">
-    <PackageVersion Include="Microsoft.Testing.Extensions.CodeCoverage" Version="18.11.0" />
+    <PackageVersion Include="Microsoft.Testing.Extensions.CodeCoverage" Version="18.11.2" />
+    <PackageVersion Include="Microsoft.Testing.Platform" Version="$(MicrosoftTestingPlatformVersion)" />
```
AArnott, line 47: What are these two _new_ `PackageVersion` items for?

nuget.config
```diff
@@ -8,6 +8,8 @@
     <clear />
     <add key="nuget" value="https://api.nuget.org/v3/index.json" />
     <add key="Consumption" value="https://pkgs.dev.azure.com/andrewarnott/OSS/_packaging/Consumption/nuget/v3/index.json" />
+    <!--<add key="test-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/test-tools/nuget/v3/index.json" />-->
```
AArnott: Can you revert all the changes to nuget.config?

Directory.Packages.props
fhnaseer: This is for compatibility. It is possible to have something like this,
mtp.cc requires Microsoft.Testing.Platform (mtp) 2.5,
xunit.mtp requires 2.0,

It's good to have test project direct reference of mtp to avoid failures when upgrading mtp.cc.

AArnott: I see. Makes sense. Although we have transitive pinning enabled so you don't actually need explicit `PackageReference` items in the test projects for this `PackageVersion` to take effect.

fhnaseer: updated,

nuget.config
fhnaseer: reverted,

